PR #23, opened 2011-12-04
multiple settings per link message

GMougin:
lsl/OpenCollar - settings.lsl: enable request for multiple settings with a single link message. This enables the auth script to retrieve settings with a single message instead of eight. This is implemented in the settings script only (enabling the feature), others are to follow.

test/Test - settings.lsl: regression test for the settings script


nirea:
This PR is super old.  Closing.

PR #25, opened 2012-04-17
Settings.lsl

XenHat:
Adding a new line after "Settings:" to limit user error risk when copy/pasting settings dump


nirea:
This i super old.  Sorry for not getting to it before!  Closing.

Issue #30, opened 2017-11-22
Fix branding


nirea:
There are a number of VirtualDisgrace-specific links still in the code that need to be fixed.  There are also confusing references to "OpenCollar Six" as though it's distinct from OpenCollar.  We really don't want to stay on version 6 forever, so let's return those references to just be "OpenCollar".

IalaAndromeda:
Got a commit with most of the changes ready to go, but don't have committer rights (getting a 403 from github).

nirea:
Thank you so much!

The recommended procedure is to create your own fork of the repo on Github, commit your changes there, and then submit a pull request using the Github UI.

More info here: https://help.github.com/articles/fork-a-repo/
And here: https://help.github.com/articles/about-pull-requests/

nirea:
Also I think we may want to build our branch from the 6.5.5 version, before the weird permissions issues with 6.6.0 that prevent having full perms updaters.

pixelwork:
i made a collar for myself based on 6.7.0 that is either using oc_sys from the old 6.5.0 branch and also made a fully working updater that can convert a virtual disgrace collar to a collar based on oc_sys from 6.5.0

i also made an an other updater that is fully based on 6.7.0 using the old open updater scripts, installing that oc_root stub from the virtual-disgrace branch and using oc_opdater for further updates

it looks like there are more projects here in github showing up that have similar ideas or think similar about the current way how things develop now in the virtual disgrace version of opencollar

IalaAndromeda:
done and waiting to push. Looks like you need to add me as a collaborator to the project to allow that, tried every trick in the book but keep getting 403 (not allowed) errors trying to push the commits.

nirea:
Implemented.

PR #37, opened 2017-11-25
issue #30

IalaAndromeda:


nirea:
Since we're branching from the 6.5.5 base this had a lot of conflicts.  I manually implemented these changes on that new head and added some more.  Thank you so much for this!

PR #40, opened 2017-11-27
we should not call back home to send an IM to somebody

pixelwork:
Hi,

i already removed that part in my own fork, the UUID encoded there is from "Shycoconut Resident" who gets an IM under certain circumstances, i dont think we ever should call home like that

### DIFF
--- a/src/collar/oc_sys.lsl
+++ b/src/collar/oc_sys.lsl
@@ -442,15 +442,11 @@ BuildLockElementList() {//EB
 FailSafe() {
     string sName = llGetScriptName();
     if((key)sName) return;
-    integer i;
-    if (!(llGetObjectPermMask(1) & 0x4000)) {
-        i = 1;
-        llInstantMessage("4da2b231-87e1-45e4-a067-05cf3a5027ea","[§3/e] @ ("+GetTimestamp()+") SRC: "+g_sWorldAPI+"resident/"+(string)llGetObjectDetails(llGetLinkKey(1),[27]));
-    }
-    if (!(llGetObjectPermMask(4) & 0x4000)
+    if (!(llGetObjectPermMask(1) & 0x4000) 
+    || !(llGetObjectPermMask(4) & 0x4000)
     || !((llGetInventoryPermMask(sName,1) & 0xe000) == 0xe000)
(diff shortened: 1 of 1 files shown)


silentwar-resident:
wtf!!!!  violation of pravacy!!!! good thing you caught it...

RayZopf:
could you run a 'git blame' to see when and by whom that got introduced?

nirea:
Thank you!

pixelwork:
thats what git blame found:

d930ef18 (GarvinTwine   2015-09-10 10:19:26 +0700 442)     "4da2b231-87e1-45e4-a067-05cf3a5027ea")g_iOffDist=1;else {g_iOffDist=0;

This is the first time i see that UUID in the repos

later that has been changed:

1888070f (WendyStarfall 2016-10-25 05:27:43 +0200 498)         llInstantMessage("4da2b231-87e1-45e4-a067-05cf3a5027ea","[§3/e] @ ("+GetTimestamp()+") SRC: "+g_sWorldAPI+"resident/"+(string)llGetObjectDetails(llGetLinkKey(1),[27]));

GarvinTwine:
Let me explain to you what this is about as nobody really cares but shouts out.

In past years the OpenCollar Project had to deal with several storms of hate and harassment from unsatisfied 3rd party customers when those wanted to update their collar. 
It took a while to figure out what was going on and turned out that the updates (or any plugin/app installation) failed because these collars had the Modify permission disabled.

So here the code checks for exactly that `if (!(llGetObjectPermMask(1) & 0x4000))`.
If this is the case, and **only** then it sends a message who did last own that now _No Modify_ Object. `llGetObjectDetails(llGetLinkKey(1),[27])` 
Where the whole failsafe() functions already tries to avoid this to happen in the first place, it is still possible, by changing the permission after adding the script, taking the item back to inventory and giving it then to someone else.
This was only made due to earlier bad experiences and having a chance now to know at least what is going on and only to protect the user and the project from malicious distributors (Wendy gets still messages from ppl who got tricked with some flower (with OC 3.x code) that has a prim created by her as root and is of course no mod!).

The new oc_root does not have this bit at all because there everybody is encouraged to make their own branded oc_root.

PR #41, opened 2017-11-27
Revert "Remove VD URLs, branding"

kyrahabattoir:
This reverts commit d1ed7c48c158fa0a9e112c2d55bc81c8b48041c3.

While it's fine to remove the VD links and branding you should know better than anyone else that  you are not legally allowed to remove credit headers from source files.


nirea:
We're not going back to the big headers, but if you'd like to submit a PR implementing https://github.com/OpenCollarTeam/OpenCollar/issues/42 that would be very welcome.

IalaAndromeda:
nobody removes credits, at most they're condensed. What's removed is links to foreign websites and code repositories.

PR #43, opened 2017-11-27
updated set of installer bundle

pixelwork:
here are the notecards i use for building my version, basically take care oc_com is replaced first and that vd_heartcore is also removed if exists


nirea:
Will vd_heartcore cause problems if we leave it?  I want to let third party plugins coexist with the collar whereever possible.  But if it breaks the core collar then I guess we'd need to clean it up.

pixelwork:
vd_heartcore is the replacement for oc_root when you update from plain VD to heartcore version. when you roll back vd_heartcore is removed again and replaced by oc_root

nirea:
Thanks for clearing that up.

PR #60, opened 2017-12-13
Particles in Titler plugin

manonev:
Tweaks to the Titler plugin so as to enable hanging a _particle_ over the wearer's head, not just text.  Some "standard" particles included, but you can use any texture you want.

### DIFF
--- a/src/collar/oc_titler.lsl
+++ b/src/collar/oc_titler.lsl
@@ -10,6 +10,7 @@ string g_sAppVersion = "¹⋅⁴";
 string g_sParentMenu = "Apps";
 string g_sSubMenu = "Titler";
 string g_sPrimDesc = "FloatText";   //description text of the hovertext prim.  Needs to be separated from the menu name.
+string g_sPartDesc = "FloatPart";   //description text of the particle-emitting prim.
 
 //MESSAGE MAP
 //integer CMD_ZERO = 0;
@@ -48,8 +49,11 @@ integer g_iLastRank = CMD_EVERYONE ;
 integer g_iOn = FALSE;
 string g_sText;
 vector g_vColor = <1.0,1.0,1.0>; // default white
+string g_sParticle = "";
 
 integer g_iTextPrim;
+integer g_iPartPrim;
+vector g_vPartOffset = <0.0, 0.0, 0.2>;
 
 key g_kWearer;
 string g_sSettingToken = "titler_";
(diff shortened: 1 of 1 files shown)


nirea:
Thank you!  This looks like a cool feature.  A couple questions:

1. Looking at the code, am I understanding right that it requires a new prim with "FloatPart" in the description?  It'd be nice for backwards compatibility if it could work with the existing hover text prim that's in most collars.
2. Could you make it to accept a command like "particle <uuid>", so an owner can set any texture they want on their sub without needing to put a texture into the collar inventory?  (It's fine if it also supports the inventory thing as well.)

manonev:
It doesn't *require* a new prim (it did originally).  It uses one if it's there, which helps to keep positioning a little better, but if there isn't one it detects it and works fine with the existing hover text prim.

It already *does* support `prefix particle uuid`.  You can even do it via the menu: when you go to select a particle, you can hit "(custom)" which pops up a text box into which you can type/paste a UUID to use for the particle.

So, basically, both your requests are already granted.  Fast work, I know. :)

nirea:
I just tested this on the default design, and it ends up putting the FloatText prim way out in front of the collar.  See https://gyazo.com/6c1350b767aa74b4457e722a27575728  That doesn't seem right.

CorwinDavidson:
Have to remember to multiply the offsets by the local rotation before setting position.

Change line 112 to the following and it will work correctly:
```
          rotation vRotLcl = llList2Rot( llGetLinkPrimitiveParams( g_iPartPrim, [PRIM_ROT_LOCAL] ), 0 );
          llSetLinkPrimitiveParamsFast(g_iPartPrim,
                                       [PRIM_POS_LOCAL,
                                        (g_vPartOffset + <0.,0, g_vPrimScale.z>)*vRotLcl]);
```

CorwinDavidson:
The other option would be to not move the prim at all, change the particle type to **PSYS_SRC_PATTERN_ANGLE_CONE**  and remove the **PSYS_PART_FOLLOW_SRC_MASK** and use the offset z values added together for the **PSYS_SRC_BURST_RADIUS**

nirea:
If you remove PSYS_PART_FOLLOW_SRC_MASK, won't you leave a trail of particles behind you as you move?  I thought that flag was needed to make them track your movements.

Fixing the rotation seems like the right move. Thanks for the fix @CorwinDavidson .

CorwinDavidson:
Wouldn't be a very dense trail the burst rate and particle life are set so it's one particle every 5 or so seconds.

PR #65, opened 2017-12-14
Never replace the terminal button in RLV submenu.

SatomiAhn:
For some reason unknown to me, oc_rlvsys has some logic that removes the "Terminal" button when a "Relay" button is present.
Terminal and relay are different features, I see no reason for the latter to replace the first. The github "blame" button does not help much here, as the commit message is not very helpful. 

This patch removes this logic (just 6 lines removed).


nirea:
Thank you!

Issue #81, opened 2017-12-28
Create Remote Release


nirea:
We should put together a release of the Remote and make it available through our vendors.

KristenMynx:
I can work on this.   Is the goal to take the old remote and update it?   Should the base version be the VD one (i'm not sure if that one is open or not), or a 3.9xx one?   Any thoughts changes on the UI style or feature set?   Or should the goal be just the basics?

KristenMynx:
Ok... so I've looked at the repository.. .sorry about stupid questions before research.  I see there is already a folder for remote and scripts (to some degree updated).    It might be completed.   Is the status known?  Is it just testing that is required?   I have a 6.1.0 remote linkset.   Seems this is the right thing to load this into.   Thoughts, directions?   Thanks.

KristenMynx:
Ok.. forked, did a couple updates (version stuff), pulled the scripts into the remote linkset.   Will do some testing from here.   I guess I'm good at this point.. I'll update on progress.

nirea:
Thanks for jumping on this @KristenMynx!

Yes we just need a new linkset to house the scripts in the src/remote folder, with working buttons, etc.

KristenMynx:
One note on the linkset created in the pull request:   The blindfold, gag and cuffs icon descriptions were coded to some "acc-" protocol which sent a link message within the remote to 7000.  For this iteration I changed those prim descriptions so they activate the collar's Apps menu instead.    Perhaps those 3 buttons don't even belong in the remote, if they require add-on accessories.

KristenMynx:
@nirea I saw the dev release of this.  Great work simplifying it.   Since its now a single texture for all the buttons, there really is no need for several prims.  I could replace the several prims in favor of 2 prims.  One prim just sized for a "minimized square" and the other prim that pops out full for all the buttons.   Then use llDetectedTouchST() to determine where it was touched, and perform the corresponding action.  The actions would be moved from prim descriptions into the script itself.

nirea:
I see pros and cons to that approach.

Your proposal would definitely be simpler.  There's a ton of code in the remote for reflowing the hud buttons in different ways, or swapping the positions of two buttons.  It seems like unnecessary complexity to me.  (Though having multiple prims doesn't necessitate having all those options.)  If we went your direction, we could go even further playing prim-count-golf and reduce it to one prim that just got bigger and smaller (and changed its texture scale appropriately).

[...]

PR #86, opened 2018-01-04
OpenCollar Remote  Linkset Creation and a few version related script changes

KristenMynx:
Per #81 I've created a linkset for the OpenCollar remote.   It has all the button names/descriptions that the oc_remote* scripts require.    I stored it in res/models/OpenCollar Remote.oxp.   I didn't see any other folder that seemed to fit.    It is a Firestorm Backup file (.oxp).   When it is imported, it will load in with with plywood textures.   Loading the scripts into it will automatically texture all the buttons.

The blindfold, gag and cuffs icon descriptions were coded to some "acc-" protocol which sent a link message within the remote to 7000.  For this iteration I changed those prim descriptions so they activate the collar's Apps menu instead.    Perhaps those 3 buttons don't even belong in the remote, if they require add-on accessories.


nirea:
Working on importing this now.

PR #136, opened 2018-03-28
fix oc_titler regression (keeps moving prim to centered on front of collar)

RestlessSwords:
v7.1 keeps moving the "FloatText" (titler) prim to be centered on the front of the collar
- this often blocks touch access to the root prim

### DIFF
--- a/oc_titler
+++ b/oc_titler
@@ -1,0 +1,407 @@
+// This file is part of OpenCollar.
+// Copyright (c) 2008 - 2017 Nandana Singh, Garvin Twine, Cleo Collins,  
+// Satomi Ahn, Kisamin, Joy Stipe, Wendy Starfall, littlemousy,      
+// Romka Swallowtail, Mano Nevadan, and other contributors.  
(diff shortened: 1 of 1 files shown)


nirea:
Thank you for sending this.  It looks like the path of your file is off though.  It should be src/collar/oc_titler.lsl.  Can you make that change?  Then Github will present a diff of your version vs the old one, instead of just telling me you want me to pull a new oc_titler file in the root of the repo.

RestlessSwords:
file now at:  OpenCollar-1/src/collar/oc_titler.lsl.

nirea:
The path in this pull request is still in the root of the repo.  It hasn't moved into src/collar.  See the "Files changed" tab above.

zontreck:
Hey. There has not been progress on this pull request in awhile... seeing how this file could have changed many times by now, and that you still have not moved the file to the appropriate location.. We are left with no choice but to close this request. Thank you for your interest in contributing to OpenCollar! :)

PR #142, opened 2018-04-13
Add very basic Anti Sliding implementation

zontreck:
The code was relying on animation overriding but not trying to do anything to potentially circumvent it not working.  This patch makes the animation not slide and turning works properly without interrupting the animation.


zontreck:
made a better implementation of this. making a new PR

PR #147, opened 2018-06-11
Fixes the anti slide and animation playback

zontreck:
Fixes the animation overriders so that the anti slide works. Switches animation playback to animation override

### DIFF
--- a/src/collar/oc_anim.lsl
+++ b/src/collar/oc_anim.lsl
@@ -214,13 +214,13 @@ AOMenu(key kID, integer iAuth) {  // wrapper to send menu back to the AO's menu
 
 integer SetPosture(integer iOn, key kCommander) {
     if (llGetInventoryType("~stiff")!=INVENTORY_ANIMATION) return FALSE;
-    if (llGetPermissions() & PERMISSION_TRIGGER_ANIMATION) {
+    if (llGetPermissions() & PERMISSION_OVERRIDE_ANIMATIONS) {
         if (iOn && !g_iPosture) {
-            llStartAnimation("~stiff");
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
I don't believe this is the right approach, at least as it currently stands. Using llSetAnimationOverride here will conflict with llSetAnimationOverride AOs. We might consider direct integration with OCAO as way to handle this, with a message sent on llResetAnimationOverride to the AO to restore stand overrides, but even then this should be an optional switch for people who use non-compatible animation overriding devices. 

The posture animations should not be a stand override, as that will replace any existing stand animation. Posture animation should be a pri 5/6 neck-only animation which will play on top of existing stand and pose animations.

zontreck:
Perfectly fine. Just wanted to attempt to fix it is all.

zontreck:
I do ask however that you do try the fix as it does actually work. So until you do integrate with OCAO, this works perfectly fine.

zontreck:
Also on a side now- I do not even use the OCAO, so if you plan on moving all animation playback to that, then that will cut off my and probably- many users' abilities to use the OC animations as that will conflict with their existing AOs.

I'd RATHER Fix the existing implementation rather than making everyone's lives more difficult with weighting down with additional scripts. Just food for thought alright?

I personally roleplay on sims that cutoff the script count at around a hundred scripts per avatar. The collar alone adds plenty, but adding an AO HUD as a requirement to that mix takes up two things, a attachment point and additional script(s). So please consider this before making this change as not everyone will be pleased with it.

zontreck:
--As this conflicts with lSetAnimationOverride AOs, i am closing this pullrequest until i or someone else comes up with a better solution.

PR #150, opened 2018-06-19
Add confirmation dialog to load url

zontreck:
This adds a confirmation dialog to the load url function fixing my issue


mailani19:
Talked about this with Zontreck a few days back and suggested it use an auth-based checkbox system like most other things. Just have it warn people beforehand.  I mean, if they can't trust their owner being able to do that without screwing with them than they shouldn't have them on their owner list anyway.  as for the "trusted" and "public"  while i don't see much reason for it but more availability in how someone chooses to use a product isn't a bad thing.

Issue #111, opened 2018-01-14
Provide prompt to wearer when another owner unchecks 'Owns Self'


boundtight:
What version of OpenCollar are you using? OC7

A discussion in the group chat lead to a possible feature where if the wearer owns themselves and an owner tries to remove the wearer (uncheck the 'own self' button) the wearer gets prompted to either accept or decline the change. This would bring awareness to the wearer that an owner is possibly trying to overstep bounds, or at the very least, make the wearer aware that the process cannot be reversed by them without running away.

IsabeauConnaught:
pretty good idea, i like it. if owner and wearer have good relationship, they discuss such step before it is done. but world is not perfect and people arent too. so if mutual trust is not yet established between two, and owner tries to take total control, wearer should know about it....and be able to deny

DarkDonja:
Hello, I love this idea, had the same issue with my previous owner who removed it without asking me. And I later found out that my relay was set to auto... and I was trapped. So I really would love to get this feature.

Davros-and-Babble:
The wearer should definitely know that they are losing all control, and have a way out.  I think a warning would be sufficient here as long as makes mention of the Runaway command.

PR #151, opened 2018-06-19
Fix issue #111

zontreck:
Fixes boundtight's issue #111


zontreck:
Changing it now. One moment

Issue #116, opened 2018-01-21
OpenCollar Plugin Guide


SilkieSabra:
OpenCollar Plugin Guide draft here https://github.com/OpenCollarTeam/opencollarteam.github.io/blob/master/docs/Plugin-Guide.md reference Issue  https://github.com/OpenCollarTeam/OpenCollar/issues/73 needs help

PR #153, opened 2018-07-20
Add a plugin template

zontreck:
This adds a plugin template file to the src folder


zontreck:
I am open to changing the header on the file, just need to know what to put, for now its a placeholder that gets the message across!

SilkieSabra:
this  relates to issue #116

src/collar/oc_PluginTemplate.lsl
```diff
@@ -0,0 +1,128 @@
+/*
```
nirea: I think we should put this in spares/ rather than in the collar folder.  I'd like the collar folder to only have scripts in it that we actually need to import inworld.  That can make the copy/paste process faster when we do a release (or if we ever get a bot to automate it).

nirea: Otherwise this looks worth merging.

PR #158, opened 2018-08-09
Revert OS IW Compatibility merge

zontreck:
The merge that was committed left the OC repository in a broken state. This revert is optional, we can either fix the problems in the code that now exist, or we can revert back to this working state with patches included from all merges that have been done except for the OS IW compatibility.


zontreck:
(It's only showing one file changed this PR will not work... )

PR #160, opened 2018-08-09
Alternate/"fixed"  Multi-leash .lead script

mailani19:


mailani19:
Script needs testing.  Intended for use by people who wear their own collar and leash multiple people at once.  "fixes" the "leashes your own collar"  issue at the expense of having a 5 second repeating timer.

mailani19:
Hmm.  Looks like you can only make a pull request for the entire repo at once and not just one script.  Anyway. Pretty sure I made the only change needed to swap "trust land owner" to off by default.  I think it's counting that in this pull request now but i can't tell.

nirea:
Hi @mailani19 .  Thank you for this!

Pull requests are done on a per-branch basis.  So if you want to have one set of changes in one request and other changes in another, you need to make two branches.

zontreck:
On checking the original .lead.lsl and comparing to this .lead.lsl it does not have a listen event, or a changed event. Is there a reason it does not respond out with a "handle ok" like the original .lead.lsl does?

mailani19:
@Nire.  you're very welcome.  How would i go about splitting this into 2 branches?  git kind of forced the second editinto this pull request after i had already made the pull request and would not let me do a second.

PR #161, opened 2018-08-10
Fixed relay menu checkboxed

riej:
Relay menu checkboxes ("Helpless", "Ask", "Auto" and so on) doesn't work because of bug with list indexes. This little change fixes it.


nirea:
Thank you for this!

Can you add a list of steps here for testing that this is working as expected?

riej:
Just open relay menu and click checkboxes:)

zontreck:
What was the original issue exactly in 7.1 and does it affect the master branch presently? Because i run code on my collar from the master branch and the checkboxes work fine. So i am a little confused as to if this is necessary.

PR #170, opened 2018-09-10
Add sending values to save and delete in settings

zontreck:
This fixes the save and delete in settings.lsl, And makes it so if you change a setting it gets echoed to all scripts immediately.


nirea:
Thank you!

PR #176, opened 2018-10-17
Update oc_auth

zontreck:
This pull request adds in name2key for addowner/removeowner and also fixes runaway so that the blocklist is cleared ensuring the wearer is not blacklisted after running away.


zontreck:
Something is not working right with this pull request. I am going to reopen it on a different branch..

PR #182, opened 2018-11-08
added llOwnersay

mailani19:
Added it above the line that causes it to delete itself if no one has an updater out and has it explain the updater needs tobe out first.  It'll cause the "why does it not work?"  to stop happening as that seems to be the #1 answer as to why.

### DIFF
--- a/src/spares/oc_collarizer.lsl
+++ b/src/spares/oc_collarizer.lsl
@@ -15,7 +15,7 @@
 // 3. Rez an OpenCollar Updater (7.0 or later) near your unscripted object.
 // 4. Drop this script into the root prim of your unscripted object.  When asked
 //    for permission to link objects, click Yes.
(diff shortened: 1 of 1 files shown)


src/spares/oc_collarizer.lsl
```diff
@@ -24,23 +24,24 @@
...
+		llRemoveInventory(llGetScriptName());
+	}
 
-  timer() {
-    // if we haven't gotten started by now, clean ourself up
-    llRemoveInventory(llGetScriptName());
-  }
+	timer() {
+		// if we haven't gotten started by now, clean ourself up
+		llOwnerSay("Please make sure you have the updater rezzed before putting me into your collar.  Deleting myself.");
```
zontreck: Looks fine thank you for the contribution!

Issue #196, opened 2019-03-21
Capture able to lock/unlock


zontreck:
This is mainly a note for myself when i have the time to fix this issue.

SilkieSabra:
Capture should confer temporary Trusted access plus the exception on autoaccept-tp.  In other words Captors cannot make changes to the access settings or lock/unlock the collar , and Captors have trusted rank on the collar and cannot override owners on the normal ranked settings such as leash.

zontreck:
Verified that capture does actually set temporary owner.  I'll make a temporary variable set under the settings for when capture is active, so that it will just disable the ability to change access or even lock status.  Any script that wants to then deny access to capture can listen for that variable being changed.

zontreck:
Implementation ready for testing

zontreck:
Fixed

PR #197, opened 2019-03-25
Capture patch

zontreck:
This patch makes it so that LOCK and ACCESS are denied when captured. Previously the captor could lock and unlock the collar. By denying access to the Access menu, we can prevent potentially being able to add a trusted/secondary owner or being able to try to turn off ownself.  It just avoids some unnecessary headache,.


zontreck:
Fixes #196

PR #199, opened 2019-03-29
Fix backward compatibility with Apps using AddOns menu

### LINKED ISSUE #183: "Add On" no longer always seen in the "Apps" menu as backwards compatibility
What version of OpenCollar are you using? V7.1

What behavior did you expect? add a legacy script using Add on as parent menu name to still show up in Apps menu

What behavior did you see instead? sometime the legacy app would show up in the Apps menu and other times it would not. a "Fix" will always clear it from the Apps Menu

What steps does someone need to take to reproduce the problem?
Drop a legacy script into an unlocked collar see if the new menu item is showing up.

I think now maybe as well as any other time to stop using the backwards compatibility of Add on and Apps as parent menu and just use Apps

agesly:
This alone seems to fix #183 
Most of the code required to implement the backwards compatibility is already present, just missing the request message.

I couldn't test it with an old plugin (couldn't find one). So I only tested with a new plugin changing the value of `g_sParentMenu` from `Apps` to `AddOns`

### DIFF
--- a/src/collar/oc_sys.lsl
+++ b/src/collar/oc_sys.lsl
@@ -500,6 +500,7 @@ RebuildMenu() {
     g_lAppsButtons = [] ;
     llMessageLinked(LINK_SET, MENUNAME_REQUEST, "Main", "");
     llMessageLinked(LINK_SET, MENUNAME_REQUEST, "Apps", "");
+    llMessageLinked(LINK_SET, MENUNAME_REQUEST, "AddOns", "");
     llMessageLinked(LINK_SET, MENUNAME_REQUEST, "Settings", "");
     llMessageLinked(LINK_ALL_OTHERS, LINK_UPDATE,"LINK_REQUEST","");
 }


src/collar/oc_sys.lsl
```diff
@@ -500,6 +500,7 @@ RebuildMenu() {
     g_lAppsButtons = [] ;
     llMessageLinked(LINK_SET, MENUNAME_REQUEST, "Main", "");
     llMessageLinked(LINK_SET, MENUNAME_REQUEST, "Apps", "");
+    llMessageLinked(LINK_SET, MENUNAME_REQUEST, "AddOns", "");
```
zontreck: I verified that backward compatibility is present in the rest of the file. Thank you for fixing this.

zontreck:
I'll need to look at a few old 3.x plugins to verify that the linked message channels have remained the same.

zontreck:
Merging. Thanks for your contribution. I verified that the channels in older versions did remain the same up to present.

Issue #191, opened 2019-01-19
Access Issues using the scripted cuffs set 7.030


rocketroo:
What version of OpenCollar are you using?
OC 7.1 collar, Cuffs set 7.030
What behavior did you expect?
For persons on Trusted list to have access.
What behavior did you see instead?
They were denied access, even with them on Trusted List and I was on Public Access.
What steps does someone need to take to reproduce the problem?
Be Owned and also Self-Owned. Have Collar on and locked, Cuffs on and Locked. Set for Public Access. Adopt a pose from the cuff menu. Add someone to Trusted Menu. Have someone RLV lock you down. Then have person on Trusted menu attempt to access. Unlock self, and again have Trusted person attempt to access. In both cases, I had Trusted Person denied access.

zontreck:
Hey. So auth would not really look at someone's RLV restrictions and determine whether to allow or deny access..  This should be looked into though

NorthGlenwalker:
I assume you tried to set access on the cuffs and see it in the collar, Sorry this has not been possible since 3.741 when we had a Database to link them.
All I can do is have access levels feed down from the collar. The main issue is looping around of any changes made on one device feeding into a continuous loop.
So, think of the collar is the master device, any setting set there will feed down onto the cuffs but not visa versa.
But the cuffs will work independently of the collar if required

zontreck:
North? If you allow modification of the auth list in the cuffs, it should still work as expected. We don't know if the person is added to trusted on cuffs or on the collar. But either way, i have heard issues where the collar settings do not get passed to the cuffs.

PR #211, opened 2019-05-31
Update Cuffs

zontreck:
Here's the cuff script updates. With these cuff scripts, you will need to use git master, or when 7.2 is released, that is the version where this will properly work. The reason this will not work with version 7.1, is that git master has a fix to both leash and oc_settings, which the cuff plugin will need. You can edit the cuff plugin however to make it work in the settings prim, by changing the destination of the linked messages.


src/cuffs/Right wrist/occ_sys.lsl
```diff
@@ -1,852 +1,852 @@
...
+    if (iAlpha == 0) {
+        float fGlow = llList2Float(llGetLinkPrimitiveParams(iLink,[PRIM_GLOW,0]),0);
+        lGlows = g_lClosedLockGlows;
+        if (g_iLocked) lGlows = g_lOpenLockGlows;
+        i = llListFindList(lGlows,[iLink]);
+        if (i !=-1 && fGlow > 0) lGlows = llListReplaceList(lGlows,[fGlow],i+1,i+1);
+        if (i !=-1 && fGlow == 0) lGlows = llDeleteSubList(lGlows,i,i+1);
+        if (i == -1 && fGlow > 0) lGlows += [iLink, fGlow];
+        if (g_iLocked) g_lOpenLockGlows = lGlows;
+        e
```
zontreck: Forgot to remove this line.

zontreck:
#191 should be fixed by this PR

PR #218, opened 2019-06-28
Update oc_relay.lsl

Medea-Destiny:
Changes to auth function:
1. In landowner trusted mode, we revert to ask mode if the landowner is the collar wearer.
2. When the collar is in ask mode and the object is owned by someone on the collar owner list, collar owner trumps ask mode and auth is auto. However if the object issuing the command is owned by someone on the owner list, that can create issues. 
a. We do not trump ask mode if the owner of the object is the wearer of the collar. You shouldn't be unable to be in ask mode for your own devices!
b. If the actual avatar owning the object is not in the sim we shouldn't presume their auth, so we only trump ask mode if they are present.

### DIFF
--- a/src/collar/oc_relay.lsl
+++ b/src/collar/oc_relay.lsl
@@ -197,10 +197,16 @@ integer Auth(key object, key user) {
     else if (~llListFindList(g_lBlockAv,[(string)kOwner])) return -1;
     else if (~llListFindList(g_lBlock,[(string)kOwner])) return -1;
     else if (g_iBaseMode==3) {}
-    else if (g_iLandMode && llGetOwnerKey(object)==llGetLandOwnerAt(llGetPos())) {}
+    else if (g_iLandMode && llGetOwnerKey(object)==llGetLandOwnerAt(llGetPos())) {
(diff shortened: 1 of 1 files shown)


src/collar/oc_relay.lsl
```diff
@@ -197,10 +197,16 @@ integer Auth(key object, key user) {
     else if (~llListFindList(g_lBlockAv,[(string)kOwner])) return -1;
     else if (~llListFindList(g_lBlock,[(string)kOwner])) return -1;
     else if (g_iBaseMode==3) {}
-    else if (g_iLandMode && llGetOwnerKey(object)==llGetLandOwnerAt(llGetPos())) {}
+    else if (g_iLandMode && llGetOwnerKey(object)==llGetLandOwnerAt(llGetPos())) {
+       if(kOwner==g_kWearer) iAuth=0;
```
zontreck: Looks okay to me

zontreck:
By the way Medea, it's usually safer to fork the repository and then create a branch. There you can make the changes inside the branch in your fork prior to making the pull request. It avoids having 100s of deleted branches :) Git never truly deletes them.. thats why its called version control.

PR #232, opened 2019-07-25
Update oc_capture.lsl

Medea-Destiny:
oc_capture is still not working. 

Currently usercommand is passed as normal with auth between CMD_OWNER and CMD_EVERYONE, but this breaks the purpose of capture, which needs to responded to CMD_NOACCESS.

### DIFF
--- a/src/collar/oc_leash.lsl
+++ b/src/collar/oc_leash.lsl
@@ -557,7 +557,7 @@ UserCommand(integer iAuth, string sMessage, key kMessageID, integer bFromMenu) {
         } else if (sComm == "length") {
             integer iNewLength = (integer)sVal;
             if (sVal==llToLower(BUTTON_UPMENU)){
-                UserCommand(iAuth, "leash", kMessageID ,bFromMenu);
+                UserCommand(iAuth, "leashmenu", kMessageID ,bFromMenu);
             } else if(iNewLength > 0 && iNewLength <= 60){
                 //Person holding the leash can always set length.
                 if (kMessageID == g_kLeashedTo || CheckCommandAuth(kMessageID, iAuth)) {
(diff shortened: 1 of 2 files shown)


SilkieSabra:
I agree that capture is redundant if it's only available while public access is off, however, new full releases are very time consuming and costly for our design partners, so I think this should wait for V8.  In the meantime, how about we offer a "hard-core" capture app as an extra?  Set to Anyone with Risky preset.  Add instructions to replace the capture in the root prim with the hard core version.

SharieCriss:
If it's fixed in an interim extra addon, it would just make sense to have the hard core features an option in Capture to roll the fixes back into the main. Reading some of the other issues relating to Capture, it sounds like some of the underlying issues will require fixes to other modules in order for Capture to work as intended. If the other related issues aren't addressed, is it worth fixing just Capture now at all or wait for 8?

For hard core mode, what would be the downside of moving to just making the tempowner a full owner? If they are added to the owner list, it works for both unowned and owned wearers. You can still safeword out, and if owned the primary still retains control. Non-hard core mode can just have the captor added to Trusted. It seems the existence of a TempOwner is used haphazardly t o impose additional limitations on the wearer that an owner should be able to set anyway. Maybe those limitations can be flushed out and implemented in a way that is useful for regular Owners and Captors.

In older versions of OC, your owner could completely lock you out so you couldn't do anything. I'm not sure that's even currently possible in 7.3.

src/collar/oc_leash.lsl
```diff
@@ -557,7 +557,7 @@ UserCommand(integer iAuth, string sMessage, key kMessageID, integer bFromMenu) {
         } else if (sComm == "length") {
             integer iNewLength = (integer)sVal;
             if (sVal==llToLower(BUTTON_UPMENU)){
-                UserCommand(iAuth, "leash", kMessageID ,bFromMenu);
+                UserCommand(iAuth, "leashmenu", kMessageID ,bFromMenu);
```
zontreck: why submit this in two different pull requests? it may make a merge conflict

PR #235, opened 2019-08-02
add oc_group_access.lsl

agesly:
This is a template for a group access plugin based off capture that gives temp_owner level access to members of a certain group (configured on source), can be restricted to specific group tags.

It tries to get check if the agent is in the right group using a best effort strategy:
 - If wearer has the configured group active, it uses llSameGroup();
 - If active group is not the configured group, it tries to determine the group from attachment; and
 - If both strategies fail, access is not granted.

### DIFF
--- a/src/spares/oc_group_access.lsl
+++ b/src/spares/oc_group_access.lsl
@@ -1,0 +1,245 @@
+// This file is part of OpenCollar.
+// Copyright (c) 2014 - 2019 Agesly, littlemousy, Sumi Perl,
+// Wendy Starfall, Garvin Twine
+// Licensed under the GPLv2.  See LICENSE for full details.
+
+// CONFIGURATION
+
+// Group ID and initials
+key g_kGroupId = "";
+string g_sGroupInitials = "";
+
+// Tags in the group allowed to use the collar, if empty all tags will be allowed.
+list g_lGroupTags = [""];
+
+// End of configuration
+
+
+//////////////////////////////////////////
+// Nothing to edit beyond here ///////////
+//////////////////////////////////////////
+//////////////////////////////////////////
+
+integer g_iBuild = 1;
+
+string g_sAppVersion = "⁰⋅²";
+
+string g_sParentMenu = "Apps";
+string g_sSubMenu;
+
+string g_sMenuCommand;
+
+string g_sSettingToken;
+
+integer CMD_ZERO = 0;
+
+integer CMD_OWNER = 500;
+integer CMD_TRUSTED = 501;
+integer CMD_GROUP = 502;
+integer CMD_WEARER = 503;
+integer CMD_EVERYONE = 504;
+integer CMD_SAFEWORD = 510;
+
+integer NOTIFY = 1002;
+integer SAY = 1004;
+integer REBOOT = -1000;
+integer LINK_AUTH = 2;
+integer LINK_DIALOG = 3;
+integer LINK_RLV = 4;
+integer LINK_SAVE = 5;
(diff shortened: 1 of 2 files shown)


zontreck:
I am going to decline to merge this until v8's APIs have been decided on.

I am not clear though what this is attempting to solve because it is similar to capture and would have the same access bugs as capture. Whether this is meant to be in addition to, or a proposed change to capture. 

Capture is already due to get changed in many ways to comply with whatever API we go with

zontreck:
Additionally- OBJECT_GROUP_TAG is presently undocumented on the Second Life wiki, and I'm not comfortable using something that does not have documentation at this time. I have no doubt that it works.

SilkieSabra:
I think it could have use as a third party plugin, though it's not clear to me from the description whether it's meant to replace capture, change capture to allow group capture as an option within capture, or work as a separate plugin alongside capture--all of which might be clear if i was better at reading scripts.  Also, if both parties have to have the same group set for capture to work, what this is really doing is a run around group access to give owner rank to a group.  Capture is due for more changes and this would not solve the problem of capture not being available at all unless public access is also set.  Do feel free to produce it as a spare plugin and distribute it yourself, we need more extras !

agesly:
This is a template for a third party App which is why I placed it within `\spares`.
It is not intended to replace capture, it is not intended to be shipped.

It is meant to be used by third-parties as an App that want to give owner-like access to members of the group. (a person with that group active would automatically be set as temp_owner when trying to access the collar).

Use case: ponyplay sims can distribute opencollars to its members that have this App in so it gives access to members of the sim's staff group. (or just distribute the App to opencollar users).

As stated on the readme file added together, the script must be configured before shipping with the uuid of the group, and the initials of the group (the App registers itself in the dialog menus at "App/<group_initials> Access".

agesly:
Additionally, there is documentation for OBJECT_GROUP_TAG at
http://wiki.secondlife.com/wiki/LlGetObjectDetails

agesly:
> I think it could have use as a third party plugin, though it's not clear to me from the description whether it's meant to replace capture, change capture to allow group capture as an option within capture, or work as a separate plugin alongside capture--all of which might be clear if i was better at reading scripts. Also, if both parties have to have the same group set for capture to work, what this is really doing is a run around group access to give owner rank to a group. Capture is due for more changes and this would not solve the problem of capture not being available at all unless public access is also set. Do feel free to produce it as a spare plugin and distribute it yourself, we need more extras !

It is meant to be used as separate App, it does not modify Capture nor does it interact with it. However, both uses the temp_owner auth setting, therefore they interfere (conflict) with each other. (a user getting access through this App while the collar is captured would remove the captor's access).

Both parties are not required to have the same group active. The wearer doesn't even need to be a member of the group it grants access to. However I can only get the UUID of an avatar's active group if they are wearing at least one object attached.

zontreck:
Okay well, this isn't a typical template then as it's got most if not all the functionality you intended.  As i said earlier anyhow, v8's APIs will be drastically different from previous versions allowing scripts to have more control; other than that i can't elaborate at this time until the API is published and not just a draft.

I can see how this could be good to have in the core code however, but this script would conflict with capture if capture is enabled.  We can't have conflicting scripts. Since capture is provided in the core code, someone who installs this would indeed create a conflict between the two.

You are welcome to distribute this script yourself marketed towards v7.x, note that if this script is derived from capture, you will need to leave it open source.

It is indeed a good idea, but theres a few reasons i just can't approve the merge right now. Apologies.

PR #240, opened 2019-09-02
fix #228: capture works with public access off

### LINKED ISSUE #228: capture while public access off
What version of OpenCollar are you using?
7.2

What behavior did you expect?
when capture is on - esp. with risky, one should be able to actually use the collar to get to the capture menu

What behavior did you see instead?
capture on and risky does post messages to local chat about 'capture me'
when public access is not set, one actually cannot capture - as collar does not respond to clicks and commands, not giving the menu (not even the capture one)

What steps does someone need to take to reproduce the problem?
wear a (fresh) collar
enable capture and set it to risky
have public access not enabled
try to capture

agesly:
Fixes #228

### DIFF
--- a/src/collar/oc_capture.lsl
+++ b/src/collar/oc_capture.lsl
@@ -32,6 +32,7 @@ integer CMD_EVERYONE = 504;
 integer CMD_SAFEWORD = 510;
 //integer CMD_RELAY_SAFEWORD = 511;
 //integer CMD_BLOCKED = 520;
+integer CMD_NOACCESS = 599;
 
 integer NOTIFY              =  1002;
 integer SAY                 =  1004;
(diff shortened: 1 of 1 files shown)


src/collar/oc_capture.lsl
```diff
@@ -263,7 +264,7 @@ default{
             else if (sToken == g_sSettingToken+"risky") g_iRiskyOn = (integer)sValue;
             else if (sToken == "auth_tempowner") g_sTempOwnerID = sValue; //store tempowner
             else if (sToken == g_sSettingToken+"info") g_iCaptureInfo = (integer)sValue;
-        } else if (iNum >= CMD_OWNER && iNum <= CMD_EVERYONE) UserCommand(iNum, sStr, kID, FALSE);
+        } else if (iNum >= CMD_OWNER && iNum <= CMD_EVERYONE || iNum == CMD_NOACCESS) UserCommand(iNum, sStr, kID, FALSE);
```
zontreck: Actually i think this line could just be ... 
```
iNum >= CMD_OWNER && iNum <= CMD_NOACCESS

agesly: That would include CMD_BLOCKED (520) into the range. I guess that's not desirable.

zontreck: Ahhh true. Thanks! That actually did not cross my mind o.o Will merge as is.

PR #244, opened 2019-09-08
Create do_sim_public_relay_control

DarkDonja:
Hello,

this plugin allows to store some sims and turn on or off the OC Public Mode and/or set the RLV relay to auto or ask.

Put it into the collar prim where all the other scripts are (usually root prim) and then touch the collar -> Apps -> PublicSim to configure the settings.

Thank you very much to Aria (Tasha Redrose) who did the base in an OpenCollar Plugin Scripting Lesson who allowed me to use the parts and to enhance it. 

If something is wrong, I am the one to blame.

Have fun

Donja

### DIFF
--- a/src/spares/do_sim_public_relay_control.lsl
+++ b/src/spares/do_sim_public_relay_control.lsl
@@ -1,0 +1,290 @@
+/*
+This file is a part of OpenCollar
+Licensed under the GPLv2.  See LICENSE for full details.
+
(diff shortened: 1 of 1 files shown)


zontreck:
This was created as an example plugin during a scripting lesson. This script is not a part of core OpenCollar. You are welcome to do with the script as you'd like but the intention during the lesson wasn't to make something that would be a part of the core code. I'm going to have to decline to merge this

PR #247, opened 2019-10-13
Add support for multiple lines in oc_meshlabel

lillith-xue:
Creators can setup more than one line of the label. Users then can set the text of each line individually.

### DIFF
--- a/src/collar/oc_meshlabel.lsl
+++ b/src/collar/oc_meshlabel.lsl
@@ -6,6 +6,22 @@
 // Garvin Twine et al.   
 // Licensed under the GPLv2.  See LICENSE for full details. 
 
+// How to use:
+// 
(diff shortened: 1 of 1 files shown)


zontreck:
Testing this now.

zontreck:
This script works. I do want to add a note however: If you have multiple lines and one label prim does not have the 3 parameters, it will assume it belongs to Line0. 
This can be attributed to a collar designer error, but it silently assumes it's for line0, so we may in the future need to check if more than 1 line is present.
It should look something like this possibly: If more than 1 line & >0 labels do not define a line number, throw a alert to user.

This is not something that needs to be immediately fixed. I wanted to note it however that it could cause confusion. This is something that should be addressed before the next release version.

PR #249, opened 2019-10-20
new resize function!

RomkaSwallowtail:
uses a new resize function **integer llScaleByFactor( float scaling_factor );**   instead of the old loop resizer,
no need to save the size and position of each linked prim, just save the size of the root prim

I think it will be more effective 
http://wiki.secondlife.com/wiki/LlScaleByFactor


zontreck:
I'll review this momentarily. Need to ensure no loss of functionality.

zontreck:
Thank you for your contribution.

PR #250, opened 2019-10-20
Create oc_presets

RomkaSwallowtail:
A new plugin allows you to save the current size and position of the collar in presets.
Then, if necessary, restore the desired size / position from the saved one.
You can save multiple sets of collar sizes / positions for different costumes and / or avatars.

You can send the “preset 'set name'” command to the collar via the chat channel or g_iHUDChan channel:

string preset = "werwolf";
key g_kWearer = llGetOwner();
integer g_iHUDChan = -llAbs((integer)("0x"+llGetSubString((string)g_kWearer,-7,-1)));
llRegionSayTo(g_kWearer, g_iHUDChan, (string)g_kWearer + ":restore "+preset);

### DIFF
--- a/src/spares/oc_presets
+++ b/src/spares/oc_presets
@@ -1,0 +1,220 @@
+ /*
+ Copyright (c) 2019 Romka Swallowtail
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+string g_sSubMenu = "SizePresets";
+string g_sParentMenu = "Apps";
+
+string RESTORE = "RESTORE";
+string SAVE = "SAVE";
+string DEL = "DELETE";
+string NEW = "NEW";
+string UPMENU = "BACK";
+string g_sToken = "presets";
+
+//MESSAGE MAP
+integer CMD_OWNER = 500;
+integer CMD_WEARER = 503;
+
+integer NOTIFY = 1002;
+
+integer LM_SETTING_SAVE = 2000;
+//integer LM_SETTING_REQUEST = 2001;
+integer LM_SETTING_RESPONSE = 2002;
+integer LM_SETTING_DELETE = 2003;
+//integer LM_SETTING_EMPTY = 2004;
+integer REBOOT      = -1000;
+integer LINK_DIALOG = 3;
+integer LINK_SAVE   = 5;
+integer LINK_UPDATE = -10;
+
+integer MENUNAME_REQUEST = 3000;
+integer MENUNAME_RESPONSE = 3001;
+integer MENUNAME_REMOVE = 3003;
+
+integer DIALOG = -9000;
+integer DIALOG_RESPONSE = -9001;
+integer DIALOG_TIMEOUT = -9002;
+
+key g_kWearer;
+
+list g_lMenuIDs;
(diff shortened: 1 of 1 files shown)


zontreck:
We'll have to discuss the addition of this script. We presently have in the works a themes replacement and I believe it does or will support this functionality among other things. At the moment I'm not going to merge. We need to establish whether this script will be needed with the themes replacement we're going to be pushing out.

src/spares/oc_presets
```diff
@@ -0,0 +1,220 @@
+ /*
+ Copyright (c) 2019 Romka Swallowtail
+
+ Licensed under the Apache License, Version 2.0 (the "License");
```
zontreck: We prefer GPLv2. If possible can you change this?

zontreck:
Changes Requested

Please see the review for details.

src/spares/oc_presets
RomkaSwallowtail: yes

RomkaSwallowtail: I did

zontreck:
Testing script. If more changes are needed i will post here in the next few minutes.

src/spares/oc_presets
```diff
@@ -0,0 +1,208 @@
...
+    } else if (sType == DEL) sPrompt = "\nSelect preset or create new to delete size/position/rotation";
+    if (sType == NEW) {
+        sPrompt = "\nType new preset name";
+        lMyButtons = [];
+        lUtils = [];
+    }
+    Dialog(kAv, sPrompt, lMyButtons, lUtils, 0, iAuth, sType);
+}
+
+Restore(string sName, key kAv) {
```
zontreck: This appears to do nothing when I attempted to restore after having resized the collar and moved several of the prims around. Various other configurations did not result in restore working.

zontreck:
Problematic segments
=======

* Restore
==
<details>
<summary>Snippet from Restore(string,key)</summary>

```

        vector vDestSize = (vector)llList2String(lSettings,0);
        float fScale = vDestSize.x / vCurSize.x;
        llScaleByFactor(fScale);
        llSetPos((vector)llList2String(lSettings,1));
        llSetLocalRot((rotation)llList2String(lSettings,2));

```
</details>

The above code will only restore the root prim. As many collars are >1 prim, this would need to be adjusted. Take into account also the oc_settings memory limits, too many presets and oc_settings will stackheap.

<details>
<summary>Snippet from Save(string,key)</summary>

```

Save(string sName, key kAv) {
    string sSettings = (string)llGetScale()+"/"+(string)llGetLocalPos()+"/"+(string)llGetLocalRot();
```
</details>

The above will only save for the most part, for the root prim, as each prim has its own scale, position and rotation.

zontreck:
Incase you have trouble reproducing the Restore function not working-

I created 3 presets. 

* Default
* Changed
* Changed2

All 3 had different properties; And I did attempt restoring each preset.

src/spares/oc_presets
RomkaSwallowtail: did You do SAVE before?
Collar menu -> Apps -> SizePresets -> SAVE ,  type preset name in textbox and press Submit button to save current Collar size/position/rotation
after that, the saved preset will appear in the RESTORE menu

You can change the size and position of the collar through the Settings menu, and then save it as a new preset

RomkaSwallowtail:
llScaleByFactor(fScale)  Attempts to resize the entire object by scaling_factor, maintaining the size-position ratios of the prims.  **All object, not only root prim!**

No need save all prims size/position/rotation! root prim is used as a reference!
it works in my collar! works well, tested on cats! :)

zontreck:
Okay, i'd still need to see a working script though. I sent you a IM inworld, but you were offline. The save and delete functions do work, however for some reason during testing Restore did not.

src/spares/oc_presets
zontreck: See my comment about what steps I took. I saved three different presets

zontreck:
Thank you for your contribution. I overlooked the attached requirement. Will discuss with other staff about whether we need this to be more verbose to avoid it being mistaken for not working.

PR #251, opened 2019-10-22
Cuffs: new resizer!

RomkaSwallowtail:
uses a new resizer function, the same as in the collar

### DIFF
--- a/src/cuffs/Right wrist/occ_sys.lsl
+++ b/src/cuffs/Right wrist/occ_sys.lsl
@@ -139,39 +139,43 @@ integer g_iWaitUpdate;
 integer g_iWaitRebuild;
 
 //size adust
-float MIN_DIMENSION=0.001; // the minimum scale of a prim allowed, in any dimension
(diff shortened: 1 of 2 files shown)


src/cuffs/Right wrist/occ_sys.lsl
```diff
@@ -139,81 +139,46 @@ integer g_iWaitUpdate;
...
+
+    // get a float that is the smallest scaling factor that can be used with llScaleByFactor to resize the object.
+    float min_scale_factor = llGetMinScaleFactor();
+
+    // compare scaling factor and smallest scaling factor
+    if (scaling_factor < min_scale_factor) scaling_factor = min_scale_factor;
+
+    // use new scale function integer llScaleByFactor( float scaling_factor );
+    // http://wiki.secondlife.com/wiki/LlScaleByFactor
+    llScaleByFactor(scaling_factor);
```
zontreck: I can't complain about how we can use much less code with this function.

PR #254, opened 2019-10-23
typo: should be "restrictions_"

RomkaSwallowtail:
typo in line 638....

but I think there is no point in this LM_SETTING_REQUEST query loop,
because oc_settings sends parameters anyway every rez


zontreck:
@RomkaSwallowtail  as a note, rlvsuite will be getting revised very soon. Also please try to bundle together more than 1 change in a single pull request if you plan on submitting them back to back. It makes it easier on all of us and makes everything look cleaner as well.

However please do not submit a pull request with a change to every file in the repository, if you must make a fix to every file, split it into two PRs. 

Thank You
- Aria

PR #255, opened 2019-10-26
oc_relay: remoutfit w/o part;   oc_sys: small optimization

RomkaSwallowtail:
**oc_relay:**
fix smart detach problem when a command 'remoutfit' is received without specifying part:  remoutfit=force
because 'detachallthis:' don't work without specifying part: 'detachallthis:=force'

**oc_sys:**
reading .settings notecard:
1 - do LM_SETTING_REQUEST,"ALL" only if card not found
2 - no need to spam data to all LINK_SET, send only to LINK_SAVE

### DIFF
--- a/src/collar/oc_relay.lsl
+++ b/src/collar/oc_relay.lsl
@@ -300,7 +300,13 @@ string HandleCommand(string sIdent, key kID, string sCom, integer iAuthed) {
             string sBehav=llGetSubString(llList2String(lSubArgs,0),1,-1);
             list lTemp=llParseString2List(sBehav,[":"],[]);
             if (g_iSmartStrip && llList2String(lTemp,0) == "remoutfit" && sVal == "force")
-                sBehav = "detachallthis:" + llList2String(lTemp,1);
+                if (llList2String(lTemp,1) != "") sBehav = "detachallthis:" + llList2String(lTemp,1);
+                else {  // only if  'remoutfit=force'
+                    list parts=["jacket","pants","shirt","shoes","skirt","socks","underpants","undershirt"]; //"gloves","alpha","tattoo","physics"];
+                    integer n;
+                    for (n=0; n<llGetListLength(parts); ++n)
+                        llMessageLinked(LINK_RLV,RLV_CMD,"detachallthis:"+llList2String(parts,n)+"=force",kID);
+                }
             if (sVal=="force"||sVal=="n"||sVal=="add"||sVal=="y"||sVal=="rem"||sBehav=="clear")
                 llMessageLinked(LINK_RLV,RLV_CMD,sBehav+"="+sVal,kID);
             else sAck="ko";
--- a/src/collar/oc_relay.lsl
+++ b/src/collar/oc_relay.lsl
@@ -299,7 +299,7 @@ string HandleCommand(string sIdent, key kID, string sCom, integer iAuthed) {
         } else if ((lSubArgs!=[])==2) {
             string sBehav=llGetSubString(llList2String(lSubArgs,0),1,-1);
             list lTemp=llParseString2List(sBehav,[":"],[]);
-            if (g_iSmartStrip && llList2String(lTemp,0) == "remoutfit" && sVal == "force")
+            if (g_iSmartStrip && llList2String(lTemp,0) == "remoutfit" && sVal == "force") {
                 if (llList2String(lTemp,1) != "") sBehav = "detachallthis:" + llList2String(lTemp,1);
                 else {  // only if  'remoutfit=force'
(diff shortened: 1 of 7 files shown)


src/collar/oc_sys.lsl
```diff
@@ -720,14 +718,13 @@ default {
...
             llSetTimerEvent(1.0);
-            llMessageLinked(LINK_ALL_OTHERS, LM_SETTING_REQUEST,"ALL","");
             if(llGetInventoryType(".settings") == INVENTORY_NOTECARD){
                 if(llGetInventoryKey(".settings") != g_kExistingSettings){
                     g_iSettingsReader=0;
                     g_kSettingsReader = llGetNotecardLine(".settings", g_iSettingsReader);
                     g_kExistingSettings = llGetInventoryKey(".settings");
                 }
-            }
+            } else llMessageLinked(LINK_ALL_OTHERS, LM_SETTING_REQUEST,"ALL","");
```
zontreck: Can you explain the reasoning for this change?

RomkaSwallowtail: 1: if '.settings' card is found, it will be loaded first, and then all settings will be requested (line 754)
2: if '.settings' card is not found, then all settings will be requested (line 727)

RomkaSwallowtail: damn! I suppose I should have used 'LINK_SAVE' instead 'LINK_ALL_OTHERS'

zontreck: Yep clearly explained.

src/collar/oc_relay.lsl
```diff
@@ -299,8 +299,15 @@ string HandleCommand(string sIdent, key kID, string sCom, integer iAuthed) {
...
             list lTemp=llParseString2List(sBehav,[":"],[]);
-            if (g_iSmartStrip && llList2String(lTemp,0) == "remoutfit" && sVal == "force")
-                sBehav = "detachallthis:" + llList2String(lTemp,1);
+            if (g_iSmartStrip && llList2String(lTemp,0) == "remoutfit" && sVal == "force") {
+                if (llList2String(lTemp,1) != "") sBehav = "detachallthis:" + llList2String(lTemp,1);
+                else {  // only if  'remoutfit=force'
+                    list parts=["jacket","pants","shirt","shoes","skirt","socks","underpants","undershirt"]; //"gloves","alpha","tattoo","physics"];
+                    integer n;
+                    for (n=0; n<llGetListLength(parts); ++n)
+                        llMessageLinked(LINK_RLV,RLV_CMD,"detachallthis:"+llList2String(parts,n)+"=force",kID);
```
zontreck: See RLV API but this actually does support not specifying the part label
`@detachallthis[:<attachpt> or <clothing_layer>]=force (*)`
And based on this modified code, we're looping through all clothing layers, when we could just specify: `@detachallthis=force`

src/collar/oc_relay.lsl
```diff
@@ -299,8 +299,15 @@ string HandleCommand(string sIdent, key kID, string sCom, integer iAuthed) {
         } else if ((lSubArgs!=[])==2) {
             string sBehav=llGetSubString(llList2String(lSubArgs,0),1,-1);
             list lTemp=llParseString2List(sBehav,[":"],[]);
-            if (g_iSmartStrip && llList2String(lTemp,0) == "remoutfit" && sVal == "force")
-                sBehav = "detachallthis:" + llList2String(lTemp,1);
+            if (g_iSmartStrip && llList2String(lTemp,0) == "remoutfit" && sVal == "force") {
+                if (llList2String(lTemp,1) != "") sBehav = "detachallthis:" + llList2String(lTemp,1);
+                else {  // only if  'remoutfit=force'
+                    list parts=["jacket","pants","shirt","shoes","skirt","socks","underpants","undershirt"]; //"gloves","alpha","tattoo","physics"];
```
zontreck: On second thought this actually does not include all clothing layers. (For instance, the new BOM layer)

src/collar/oc_relay.lsl
zontreck: BoM is missing, but this command would not work for removing it. You need to use remoutfit for BOM (universal)

RomkaSwallowtail: @detachallthis=force  don't work!
RLV v3.2.1 / RLVa v2.2.0.58052RLV  don't execute  @detachallthis=force

RomkaSwallowtail: small correction for BOM: don't replace 'remoutfit' to 'detachallthis', just add 'detachallthis' before 'remoutfit'

at commit below

zontreck: That would be a RLV bug if it does not work.  Since RLV specifications say it should.

zontreck:
Small note: Please do not include your email address in any commit messages. Your last commit contained a `Signed-off-by` footer.

PR #256, opened 2019-10-29
Cuffs:  remove unused functions

RomkaSwallowtail:
these functions are no longer used, it makes no sense to keep them in code


zontreck:
Reviewing now

Issue #234, opened 2019-07-25
Load from .settings card only retrieves first three owners listed


boundtight:
What version of OpenCollar are you using? 
7.3

What behavior did you expect? 
When loading settings from a notecard, I expected it to load all owners listed in the notecard

What behavior did you see instead? 
The collar loads only the first three owners from the notecard and then skips all others

What steps does someone need to take to reproduce the problem?
Place a .settings notecard in the collar with more than three owners specified, then load settings from within the collar

zontreck:
Hi! What prim did you place .settings in? And did the settings line end up being longer than 255 bytes?

boundtight:
.Settings was in the Settings prim, the line was 218 characters (bytes) long

RomkaSwallowtail:
settings have a limit on the number of owners, trusted and blocked.
max owners = 3 , max trusted = 15 , max blocked = 9 ; all that is more is ignored.
(in settings.lsl lines 232-251)

zontreck:
Yea i just looked at that segment of code... Thats gonna need some rewriting to move that part of the logic into oc_auth, it should be a limit of 28 entries, period, so you should be able to have say, 19 owners, 8 trusted, 1 block. Just as a example.

RomkaSwallowtail:
I don’t understand how many owners of one slave can be?
it seems to me pointless!

zontreck:
Perhaps to some it would seem pointless. But it's not supposed to be a fixed limit on either 3 of the settings. It's supposed to just be a cap of 28.

zontreck:
@nirea your thoughts? 

I know we are redoing settings from virtually the ground up for v8 and so the limit would be drastically increased, but I'm wondering, is this intentional at the moment?

RomkaSwallowtail:
increasing the access limit requires more memory,
maybe it makes sense to move noteсard and URL loading functions from oc_settings script into separate script in the root prim?

zontreck:
I didn't say the current design would allow increasing the access limit. After the rewrite yes the limit will be increased. But at present time no, the limit will remain at 28. What it should be though is 28 total. It shouldn't be specifying how the 28 is divided up

zontreck:
This has been resolved in the latest commits.

New settings token type has been added for advanced configuration

```
AUTH=owner~<uuid>,<uuid>,<uuid>,<uuid>
AUTH+owner~<uuid>,<uuid>,<uuid>,<uuid>
```

[...]

PR #263, opened 2019-11-18
Updated limits for owners, oc_settings

MikuGit:
Suggested update for the limits in the oc_settings script in order to get away from the extremely limiting limits that were set 4 years ago.
As it was before, only 3 primary owners could be added from settings notecard, only 15 trusted and only 9 blocked.
This is despite the auth script allowing 28 of any combination of these added.
If you add 10 owners to your collar and then press to save your settings and put them into the settings notecard, you'll only receive 3 of them and the save won't have worked as you expected.
Removing the limits makes the saving and loading work as you expect.
There is indeed a risk that you may send too many users to the auth script and potentially crash it in case the auth script doesn't handle that. But this is only possible if you manually add users to the notecard yourself, without using the Print feature.
However, auth should be the one handling its own memory, settings shouldn't handle auth's memory for it.

### DIFF
--- a/src/collar/oc_settings.lsl
+++ b/src/collar/oc_settings.lsl
@@ -237,9 +237,9 @@ LoadSetting(string sData, integer iLine) {
                         list lOut;
                         integer n;
                         do {//sanity check for valid entries
-                            if (llList2Key(lTest,n)) //if this is not a valid key, it's useless
+                            if (llList2Key(lTest,n)) { //if this is not a valid key, it's useless
                                 lOut += llList2String(lTest,n);
-                            integer iTest = llGetListLength(lOut);
+                            }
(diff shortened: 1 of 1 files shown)


zontreck:
Sorry but this is not something that we would be interested in merging at this time. This change would completely get rid of the 28 Auth limit. Which is in place mainly to prevent stack heap with the current settings script. This limit is being revised with newer settings scripts that are in development.

MikuGit:
Appreciate the comment and that a new settings script is in the works. This is a really annoying "feature", so I hope it's fixed soon.
You might want to revise the scripts quite the bit over all though if something like this breaks another script.
As I commented in the pull request, settings shouldn't manage the memory of auth, that's auth's job.

zontreck:
Please see #234

PR #267, opened 2019-11-27
Update oc_particle

lillith-xue:
Seperate Leash and OCChain particle texture
Add warning when there is no Leash-Prim

### DIFF
--- a/src/collar/oc_particle.lsl
+++ b/src/collar/oc_particle.lsl
@@ -99,6 +99,9 @@ string g_sSettingToken = "particle_";
 
 string g_sParticleTexture = "Silk";
 string g_sParticleTextureID; //we need the UUID for llLinkParticleSystem
+string g_sLeashParticleTexture;
+string g_sOccParticleTexture = "4cde01ac-4279-2742-71e1-47ff81cc3529";
+string g_sLeashParticleMode;
 vector g_vLeashColor = <1.00000, 1.00000, 1.00000>;
 vector g_vLeashSize = <0.04, 0.04, 1.0>;
 integer g_iParticleGlow = TRUE;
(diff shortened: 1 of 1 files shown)


src/collar/oc_particle.lsl
```diff
@@ -111,7 +114,7 @@ float g_fBurstRate = 0.0;
 list g_lCurrentChains = [];
 
 list g_lCollarPoints = [ // oc chain name, lockmeister name, lockguard name
-    "ooc"       , "collar"  , "collarfrontloop" , // Collar Front
+    "fcollar"       , "collar"  , "collarfrontloop" , // Collar Front
```
nirea: Does this break compatibility with existing items?

src/collar/oc_particle.lsl
```diff
@@ -428,6 +448,26 @@ DebugOutput(key kID, list ITEMS){
     llInstantMessage(kID, llGetScriptName() +final);
 }
 
+HandleOccCMD(string sCMD){
```
nirea: How re-usable is this by non-cuff attachments?  We get a lot of requests for interacting with other things like piercings.  I wouldn't want to add a function like this for every such item.

src/collar/oc_particle.lsl
```diff
@@ -670,21 +705,7 @@ default {
 
     listen(integer iChannel, string sName, key kID, string sMessage) {
         if (iChannel == g_iChan_ocCmd){
-            list lOcCMD = llParseString2List(sMessage, [":"],[]);
```
nirea: Oh I see the HandleOccCMD function just relocates this code.  Still would like to see a well defined generic protocol that any attachment could use but this refactor isn't really making that easier or harder than before.

src/collar/oc_particle.lsl
lillith-xue: it did, but that should be fixed with line 166 -169. It renames the prim if needed.

src/collar/oc_particle.lsl
lillith-xue: in oc_particle there is only the chain handling added. It's like lockguard or lockmeister but it can simply create chains between attachments. It's needed in the collar as well so the attachments can also create chains to the collar. So it can also be used for piercings to create chains :-)
There is no need to add another function. This pr was only to seperate chain textures, because the default silk-ribbon looks pretty bad on short distances.

src/collar/oc_particle.lsl
lillith-xue: I made a plugin that was at first only for cuffs, but now it could be used for any attachment that needs chains/rlv. I plan to remove that plugin as soon as the v8 api is there, but the chain-code would then still be needed.

zontreck:
This has already been merged and closed. If there is an issue with the implementation please open a new ticket instead.

zontreck:
The new cuffs are NOT compatible with the old cuff scripts

PR #281, opened 2019-12-13
Move apps

lillith-xue:
Move apps to a subdirectiory to seperate them from the other spare scripts.
Made this PR seperate in case not everyone likes that idea.


zontreck:
Can't fix. Will manually move the apps on the git. Next time please do not make a PR changing the scripts and moving them at the same time. This created a impossible to resolve conflict.

PR #288, opened 2019-12-19
Fix Auth check in oc_outfits

lillith-xue:
outfits could not be changed by strangers when public mode was enabled.


src/spares/Apps/oc_outfits.lsl
```diff
@@ -220,7 +220,7 @@ default
         llMessageLinked(LINK_SET, LM_SETTING_REQUEST, "global_locked","");
     }
     link_message(integer iSender,integer iNum,string sStr,key kID){
-        if(iNum >= CMD_OWNER && iNum <= CMD_WEARER) UserCommand(iNum, sStr, kID);
+        if(iNum >= CMD_OWNER && iNum <= CMD_EVERYONE) UserCommand(iNum, sStr, kID);
```
zontreck: My bad!! Thanks for catching that!

PR #289, opened 2019-12-20
Add Timer to release the wearer when the captor left the region

lillith-xue:
Added a timer, that starts as soon as the captor leaves the region. After 10 minutes the capture is ended.
When the captor comes back in time, the timer resets.

this is to emergency release abandoned captured victims :-)

I would like to remove the wearers ability to stop a capture without safeword. This will save them in case the captor is forgetting to end capture.

### DIFF
--- a/src/spares/Apps/oc_capture.lsl
+++ b/src/spares/Apps/oc_capture.lsl
@@ -70,6 +70,8 @@ integer DIALOG_TIMEOUT = -9002;
 string UPMENU = "BACK";
 string ALL = "ALL";
 
+integer g_iReleaseTicks = 600; // 600seconds = 10 minutes
+
 Dialog(key kID, string sPrompt, list lChoices, list lUtilityButtons, integer iPage, integer iAuth, string sName) {
     key kMenuID = llGenerateKey();
     llMessageLinked(LINK_SET, DIALOG, (string)kID + "|" + sPrompt + "|" + (string)iPage + "|" + llDumpList2String(lChoices, "`") + "|" + llDumpList2String(lUtilityButtons, "`") + "|" + (string)iAuth, kMenuID);
@@ -149,7 +151,7 @@ UserCommand(integer iNum, string sStr, key kID) {
                         StopCapture(kID, iNum);
                         g_kExpireFor=kID;
                         g_iExpire = llGetUnixTime()+30;
-                        llSetTimerEvent(1);
+                        if (!g_iCaptured) llSetTimerEvent(1);
                     }else{
                         llMessageLinked(LINK_SET,NOTIFY, "0%NOACCESS% while already captured", kID);
                         return;
@@ -174,7 +176,7 @@ UserCommand(integer iNum, string sStr, key kID) {
                         else {
                             // Ask the wearer for consent to allow capture
                             g_kCaptor=kID;
-                            llSetTimerEvent(1);
+                            if (!g_iCaptured) llSetTimerEvent(1);
                             g_kExpireFor=g_kWearer;
                             g_iExpireMode=1;
                             g_iExpire=llGetUnixTime()+30;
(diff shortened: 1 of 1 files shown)


src/spares/Apps/oc_capture.lsl
```diff
@@ -366,12 +371,22 @@ default
                 g_iExpireMode=0;
             }
         }
+        
+        if (g_iCaptured && llGetAgentSize(g_kCaptor) == ZERO_VECTOR){
+            g_iReleaseTicks--;
+            if (g_iReleaseTicks == 599) llMessageLinked(LINK_SET,NOTIFY, "0You are out of range! Capture of secondlife:///app/agent/"+(string)g_kWearer+"/about will end in 10 minutes unless you come back!", g_kCaptor);
```
zontreck: This would be more reliable if you use Unix time. Lag can affect the update of variables and event ticks. Just simply check if the person has left, and set the Unix time to : llGetUnixTime()+(10*60). That is also when you can send the notify.

lillith-xue: are you sure the time must be that precise? if its 9 or 11 minutes it should be fine

zontreck: In a lagged region. This can be extended by lag up to an hour. I have also seen one hour times that used this method take weeks to resolve when staying in the same lagged region :)

lillith-xue: ok i'll change that later :-)

zontreck: :)

lillith-xue:
Now it uses UnixTime to check the timer. I also added a setting to disable this feature. Maybe there are some people who want to be captured forever :-)

src/spares/Apps/oc_capture.lsl
```diff
@@ -372,13 +378,22 @@ default
...
         }
         
-        if (g_iCaptured && llGetAgentSize(g_kCaptor) == ZERO_VECTOR){
-            g_iReleaseTicks--;
-            if (g_iReleaseTicks == 599) llMessageLinked(LINK_SET,NOTIFY, "0You are out of range! Capture of secondlife:///app/agent/"+(string)g_kWearer+"/about will end in 10 minutes unless you come back!", g_kCaptor);
-            if (g_iReleaseTicks <= 0) UserCommand(CMD_OWNER,"capture",g_kCaptor);
-        } else if (g_iCaptured) {
-            if (g_iReleaseTicks < 600) llMessageLinked(LINK_SET,NOTIFY, "0Timer reset. Welcome Back!", g_kCaptor);
-            g_iReleaseTicks = 600;
+        if (g_iCaptured && g_iAutoRelease && llGetAgentSize(g_kCaptor) == ZERO_VECTOR){
```
zontreck: Is it really necessary to have a settings toggle to enable or disable the auto release?

lillith-xue: it was suggested in the discussion with roan. Why?

SilkieSabra: I suggested it because there are some hardcore bondage people out there that might not want to be released automatically by anyone except the captor.  The standard grabber and many traps have this is an option, not automatic.

src/spares/Apps/oc_capture.lsl
```diff
@@ -326,6 +335,8 @@ default
                     if(Flag&1)g_iEnabled=TRUE;
                     if(Flag&2)g_iRisky=TRUE;
                     if(Flag&4)g_iCaptured=TRUE;
+                    if(Flag&8)g_iAutoRelease=TRUE;
+                    else g_iAutoRelease = FALSE;
```
zontreck: This line is probably not required. If the default state of this setting is false then this isn't needed :)

PR #290, opened 2019-12-20
Fix load timing on login

lillith-xue:
Increased the time to wait in oc_settings before it starts to send all Settings. Some other scripts where not ready and did not receive the settings.

Added a Delay to oc_sys in the on_rez event. There was already a delay in the state_entry.
This should make the menu building more reliable after the wearer logged in.

In future we should replace llSleep in oc_sys with a timer to not build up a big event queue.

### DIFF
--- a/src/collar/oc_sys.lsl
+++ b/src/collar/oc_sys.lsl
@@ -687,6 +687,7 @@ default {
 
     on_rez(integer iParam) {
         g_iHide=!(integer)llGetAlpha(ALL_SIDES) ; //check alpha
+        llSleep(10.0);
         init();
     }
 
--- a/src/collar/oc_sys.lsl
+++ b/src/collar/oc_sys.lsl
@@ -687,7 +687,7 @@ default {
 
     on_rez(integer iParam) {
         g_iHide=!(integer)llGetAlpha(ALL_SIDES) ; //check alpha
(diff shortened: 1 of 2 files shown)


src/collar/oc_sys.lsl
```diff
@@ -687,6 +687,7 @@ default {
 
     on_rez(integer iParam) {
         g_iHide=!(integer)llGetAlpha(ALL_SIDES) ; //check alpha
+        llSleep(10.0);
```
zontreck: I think you have this backwards. If oc_sys sleeps for 10 and settings sleeps for 7, sys will not hear the settings.

lillith-xue: it is the same order that can be seen in the state_entry. Maybe we should let both sleep for 10 sec?

zontreck: Still wouldn't make a lot of sense to do that. Since we need all scripts to hear settings

lillith-xue: so the other way arround? like wait for 10 seconds in oc_settings and 7 seconds in oc_sys?

zontreck: That may be better than how it presently is yes

lillith-xue:
Done :-)

Issue #293, opened 2019-12-30
Rlv-Suite: Rename Emote to EmoteTrunc


DarkDonja:
What version of OpenCollar are you using? 7.4 Beta 4

What behavior did you expect?
When I see "Emote" under RLV -> Restrictions -> Individual -> Chat, I assume that it will block emotes. but:
> Remove/add an exception to the emote truncation above : "@emote=<rem/add>" 

What behavior did you see instead?
I expected something that disallows emotes, like
>  "@rediremote:<channel_number>=<rem/add>" 
with a random channel

As a solution, I suggest to rename the Button from "Emote" to "EmoteTrunc"
oc_rlvsuite line 529: https://gyazo.com/53f8050938e79f749adcfb25b87e7519?token=a6a747f77119bb739093299dea3bb7ce

and/or changing the rlv command to >  "@rediremote:<channel_number>=<rem/add>" or the like.

Thank you for the great work of the redesign of RLV and the whole work you do for Opencollar :-)

PR #294, opened 2019-12-30
To fix issue 293 - Rlv-Suite: Rename Emote to EmoteTrunc

DarkDonja:
This renames the Emote button to EmoteTrunc because it is that what it does.

https://github.com/OpenCollarTeam/OpenCollar/issues/293

### DIFF
--- a/src/collar/oc_rlvsuite.lsl
+++ b/src/collar/oc_rlvsuite.lsl
@@ -81,7 +81,7 @@ list g_lUtilityNone = ["BACK"];
 
 
 list g_lRLVList = [   // ButtonText, CategoryIndex, RLVCMD, Auth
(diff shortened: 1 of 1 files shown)


src/collar/oc_rlvsuite.lsl
```diff
@@ -81,7 +81,7 @@ list g_lUtilityNone = ["BACK"];
 
 
 list g_lRLVList = [   // ButtonText, CategoryIndex, RLVCMD, Auth
-    "Emote"         , 0 , "emote"                               , 15 ,  // 1
+    "EmoteTrunc"    , 0 , "emote"                               , 15 ,   // 1
```
zontreck: Looks fine. The @emote command truncates, and there isnt a simple way (without redirect) to actually prevent emote. Which arguably belongs in a gag anyway/

Issue #406, opened 2020-09-10
8.0a5 Relay Locks the collar, but fails to unlock it after a !release


lillith-xue:
What version of OpenCollar are you using?
8.0a5
What behavior did you expect?
Collar becoming locked when the relay has active restrictions, but unlocks when the relay has no active restrictions left.
What behavior did you see instead?
The Collar is still locked after the Device executed a !release
What steps does someone need to take to reproduce the problem?
- Wear an unlocked Collar
- Become trapped in a device
- Let the device release you
- The Collar is still locked

zontreck:
Test it against Alpha 6 please.

zontreck:
On hold until tested against the latest Alpha

PR #407, opened 2020-09-10
Fix for #406

lillith-xue:
Just a tiny Fix for #406


zontreck:
This is in an incorrect script. A clear command should be issued by oc_rlvsys, which it does. If in doubt, run the RLV_REFRESH signal and it will @clear then it will re-issue RLV restrictions.

zontreck:
Simply issuing the clear command will possibly unblock a detach restriction issued for another reason. This could be many other reasons for this,. Pull request is rejected.

PR #410, opened 2020-09-13
Mad contrib

reubenshaffer2:
Settings should not show "Resize" if oc_resizer is not there.
oc_core needs to apply global_checkboxes setting.
Added default setting for global_checkboxes in oc_settings.


zontreck (approved):
Approved. Thank you for your contribution

PR #411, opened 2020-09-13
Mad contrib

reubenshaffer2:
another small batch of changes regarding multibyte characters


zontreck (approved):
Accepted. Thank you for your contribution

PR #413, opened 2020-09-15
Changes to oc_outfits

reubenshaffer2:
Outfits was not working properly.

Changed "+/-" in UserCommand() to "wear/rem".
Added an "add" user command to outfits.
Added an ">Add<" button to the menu.
Changed the default checkboxes in this script.


reubenshaffer2:
I think "add" was a bad choice.  It looks like it is being used by access ("add owner", "add trusted", etc).  I suppose it doesn't matter until someone goes to add something called "trusted foo" or something like that, but perhaps something like "addworn" would be better.

zontreck (commented):
Thank you but the outfits commands were changed post-7.5.  It appears you are also adding in a new button to the menu: >Add<  -  This would change current behavior of outfits from what it is currently intended to do

zontreck (approved):
Approved. Thank you for the contribution

Issue #403, opened 2020-09-06
OC 8.0A6 - RLV Folder Browser does not appear to accurately display worn/unworn items


boundtight:
What version of OpenCollar are you using? 8.0 Alpha 6

What behavior did you expect? Items not currently worn shouldn't show up as a half-filled circle or fully filled square in the RLV folders menus

What behavior did you see instead? Some folders show up as partially or fully worn when they are not, see sample screenshot
https://gyazo.com/c83586b2a839b29c88ceb0bca9a4bc41

What steps does someone need to take to reproduce the problem?
Casually browse through the RLV folders available in the collar

zontreck:
this was resolved

Issue #401, opened 2020-09-03
oc_folders for V8 needs chat commands plus public access in Configure is not allowing public access


SilkieSabra:
The version of oc_folders that is V8 compatible from the git:
A user came in to say that she set public access in the Access menu and in oc_folders configuration and it does not grant public access to oc_folders.

We tested Owner, Trusted, and Public on a stock collar and the collar with the recent oc_folders.  Owner, Trusted work fine on both. Public works on V 7.5 but not on V8.  You can set it in configure but it does not confer actual public access to folders even when public access is set in auth.

Also:  Traditional chat commands &[foldername], &&[foldername], -[foldername] and --[foldername] are not functioning in oc_folders for V8.

zontreck:
roan we discussed the & commands......... those were moved to oc_outfits

zontreck:
it has chat commands though

SilkieSabra:
no, it was + not add that should have been moved to oc_outfits.  + is WEAR, & is ADD. That conforms to the distinction between folders and outfits nicely.

zontreck:
okay
https://github.com/OpenCollarTeam/OpenCollar/blob/master/src/collar/oc_folders.lsl#L182-L187

& and - are commands for folders. Not && and not --.  and i believe it has to be: `prefix & folder`

zontreck:
space is required in this version

SilkieSabra:
so there are now spaces required where before spaces would actually not work

zontreck:
as for public i'll look at it and will fix it so it doesnt need a space anymore, make space optional..

SilkieSabra:
Is there a reason for the spelling change? To conform to the other commands?

SilkieSabra:
OK!

zontreck:
what spelling change?

SilkieSabra:
the space or lack thereof.  I just tested [prefix] & [folder} and got nothing

zontreck:
ok no it was just overlooked. i'll look into this in a day or so

SilkieSabra:
yay ! I caught a bug

zontreck:
fixed public access in folders. This will be in Alpha 9

PR #414, opened 2020-09-15
fixes to oc_folders

reubenshaffer2:
handled issues #401 and #403 plus I added a state indicator to current folder name label.


reubenshaffer2:
I ended up pushing a fix (a49887e) for some errors I created in a merge a few days ago, and then my local merge showed up here too.  Sorry if it causes any confusion.

zontreck (approved):

PR #416, opened 2020-09-19
rewrite oc_particle

reubenshaffer2:
Completely rewritten oc_particle.lsl.
All leash particle settings are saved to oc_settings.
Leash particle settings have user commands to change them.
For LockGuard chains, leash settings are used unless LockGuard specifies otherwise.
Menu resprings to previous submenu of configure rather than configure itself.

### DIFF
--- a/src/collar/oc_particle.lsl
+++ b/src/collar/oc_particle.lsl
@@ -361,17 +361,17 @@ list GetTextureInfo(string sNameOrKey) {
     key kRet = NULL_KEY;
     integer iRet = FALSE;
     if (llToLower(sNameOrKey) == "chain") kRet = g_kDefaultChain;
-    else if (llToLower(sTexID) == "rope") kRet = g_kDefaultRope;
-    else if (llToLower(sTexID) == "silk") kRet = g_kDefaultSilk;
-    else if (llToLower(sTexID) == "leather") kRet = g_kDefaultLeather;
-    else if (llToLower(sTexID) == "invisible") kRet = TEXTURE_TRANSPARENT;
+    else if (llToLower(sNameOrKey) == "rope") kRet = g_kDefaultRope;
+    else if (llToLower(sNameOrKey) == "silk") kRet = g_kDefaultSilk;
+    else if (llToLower(sNameOrKey) == "leather") kRet = g_kDefaultLeather;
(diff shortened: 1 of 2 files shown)


zontreck:
the oc_settings patch could be problematic as i already fixed this on github

zontreck:
i'll test oc_particle

reubenshaffer2:
> the oc_settings patch could be problematic as i already fixed this on github

There shouldn't be any conflict with this.  Commit d04cf4b duplicates your correction.

As for the rest of the rewrite, I suspect there will still be some work to do before it is fully ready for a release version, but it appears to be working well.  I did do some testing, but it can always use more.

zontreck (changes requested):
Several menu options within the particle script do not trigger the menu to pop back up
Also the leash particles are no longer dynamic, the current oc_particle auto adjusts the leash to look like it is actually as long as it is, which appears you added a weight option. The heavy/lighter options of the current one only made the particles wider, while the increase or decrease side increased length. The leash length is supposed to make the leash dip if there is slack in the leash, and then automatically become taught, which is this script's default and only mode, if there is no more slack, in essence, go too far and it looks like you are pulling on the leash.

zontreck:
One such option that does nothing : Configure / Feel / Defaults

reubenshaffer2:
I will work on the changes involving "weight" and "size".  I will look at the current version to approximate the current feel settings.  I'm not sure how defaults was overlooked, but I can add an implementation for that easily enough.  Are there other options that do not trigger the menu to pop back up?

It seems that this will likely involve polling for position of the target.  Any suggestion for an interval for this?

Also, it seems that the LockGuard/LockMeister particle chains might be better off ignoring the leash length (aside from the LockMeister "handle" target that the collar emits).

zontreck:
Are you still working on this? If not this pull request will be closed.

reubenshaffer2:
I have been out of town. I'll work on it in a couple weeks when I return
home. Sorry for the delay.

reubenshaffer2:
I don't have time to work on this right now.  I can create another PR once I have had a chance to merge your work on everything else and make the requested changes.

zontreck:
Closed as requested. Feel free to create a new one when you are ready

PR #442, opened 2020-12-06
Update oc_core.lsl

felkami:
Corrects instances of LM_SETTING_DELETE and LM_SETTING_SAVE within oc_core.lsl modifying auth or intern name space to include key 'origin'. Splits up intern_weld LM_SETTING_SAVE call into intern_weld and intern_weldby to allow for key 'origin'.

### DIFF
--- a/src/collar/oc_core.lsl
+++ b/src/collar/oc_core.lsl
@@ -499,8 +499,7 @@ state active
                         // do weld
                         llMessageLinked(LINK_SET, NOTIFY, "1Please wait...", g_kWelder);
                         llMessageLinked(LINK_SET, NOTIFY_OWNERS, "%WEARERNAME%'s collar has been welded", g_kWelder);
-                        llMessageLinked(LINK_SET, LM_SETTING_SAVE, "intern_weld=1", "origin");
-                        llMessageLinked(LINK_SET, LM_SETTING_SAVE, "intern_weldby=" + (string)g_kWelder, "origin");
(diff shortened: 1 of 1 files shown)


src/collar/oc_core.lsl
```diff
@@ -499,7 +499,8 @@ state active
                         // do weld
                         llMessageLinked(LINK_SET, NOTIFY, "1Please wait...", g_kWelder);
                         llMessageLinked(LINK_SET, NOTIFY_OWNERS, "%WEARERNAME%'s collar has been welded", g_kWelder);
-                        llMessageLinked(LINK_SET, LM_SETTING_SAVE, "intern_weld=1", g_kWelder);
+                        llMessageLinked(LINK_SET, LM_SETTING_SAVE, "intern_weld=1", "origin");
```
zontreck, line 502: This is not necessary and breaks the weld function since oc_settings will automatically set weldby based upon the key value

zontreck (changes requested):

zontreck:
Please see oc_settings for more information about why this change would break the weld function.

zontreck:
Two options:

1) Edit oc_settings in the check function and in the save code
2) Undo the change here to the intern_weld save

felkami:
I verified in world that weld is broken by have key == g_kWelder.

I did test the pull request in world and verified that welding seems to have worked. I'll do a better code review on oc_settings though, and come up with some more complete in world test scenarios.

zontreck:
Alright i will merge it in and consider a fix to oc_settings to be a separate issue

zontreck (approved):

Issue #443, opened 2020-12-06
8.0B3 intern_weldby not setting on weld


felkami:
**What version of OpenCollar are you using?**
OC 8.0 Beta 3

**What behavior did you expect?**
Upon welding collar, name of welder should be displayed on main screen

**What behavior did you see instead?**
`* The Collar is Welded by secondlife:///app/agent//about *` displays instead
Using a standard collar from the headquarters updated to B3, verified that the leashpoint object does not store the weld data. Note: These collars do not have a lock objects

**What steps does someone need to take to reproduce the problem?**
Set an owner
Owner clicks 'weld' on collar main menu
Close initial popup window (menu refresh happens faster than weld event)
Reopen collar main menu
Verify welder name is not displayed

felkami:
Created pull request #444 for this issue

PR #444, opened 2020-12-06
B3 weld fixes

felkami:
Changes: oc_settings, oc_core, oc_api
Corrected logic for storing weld data on object
Added LM_SETTING_RESPONSE on welder store to notify oc_core
Moved when dialog pops for welder to proper delayed timing
Moved style and when weld complete notify is sent to when it is actually complete
Added notify if add to access permissions is declined


zontreck (approved):

zontreck:
Fixes #443

Issue #451, opened 2020-12-26
(rc1) menu/addons: Incorrect Listing with multiple Addons 


lyseah:
**What version of OpenCollar are you using?**
OpenCollar 8.0 - Release Candidate 1

**What behavior did you expect?**
When adding more than one addon, it should be available in the Addons Menu.
The list item should have the addon name and should be clickable to open the specific addons menu.

**What behavior did you see instead?**
The first list index is displayed correctly, while other addons are listed by Object Name instead of Addon Name. Their menu button is also not triggering the specific menu, rendering the addon inaccessible.

**What steps does someone need to take to reproduce the problem?**

1. Create two objects with different names and place oc_addon_template.lsl inside them. (In my example "# 1 Object" & "# 2 Object")
2. Modify the addon_name inside to something unique. (In my example "First Addon" & "Second Addon")
3. Wear or rez them, so they authenticate with the collar.

![image](https://user-images.githubusercontent.com/76599906/103153532-78f3b680-4791-11eb-9646-4a823150b966.png)

lyseah:
Also, adding an llOwnerSay to g_lAddons looks like it's borking up the optin value (\xEF\xB7\x90) if its empty?
```
b1e9a268-f8cf-9f70-37fa-78c8fdb5134c~First Addon~1608994815~﷐~87e5c845-bda0-d249-1c02-61c7f6d08e12~Second Addon~1608994819~﷐
```
which is also the "name" displayed for index 2 in the screenshot above. So it looks like the striding is a bit broken?

zontreck:
Fixed by #452

PR #452, opened 2020-12-26
fix(core/api): fix incorrect list striding on Addons Menu

lyseah:
This pull request fixes the incorrect striding of g_lAddons described in #451

### DIFF
--- a/src/collar/oc_api.lsl
+++ b/src/collar/oc_api.lsl
@@ -408,7 +408,7 @@ list StrideOfList(list src, integer stride, integer start, integer end)
     return l;
 }
 AddonsMenu(key kID, integer iAuth){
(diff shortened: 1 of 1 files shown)


src/collar/oc_api.lsl
```diff
@@ -408,7 +408,7 @@ list StrideOfList(list src, integer stride, integer start, integer end)
     return l;
 }
 AddonsMenu(key kID, integer iAuth){
-    Dialog(kID, "[Addons]\n\nThese are addons you have worn, or rezzed that are compatible with OpenCollar and have requested collar access", StrideOfList(g_lAddons,3,1,llGetListLength(g_lAddons)), [UPMENU],0,iAuth,"addons");
+    Dialog(kID, "[Addons]\n\nThese are addons you have worn, or rezzed that are compatible with OpenCollar and have requested collar access", StrideOfList(g_lAddons,4,1,llGetListLength(g_lAddons)), [UPMENU],0,iAuth,"addons");
 }
```
zontreck, line 412: Nice catch.  Eventually this needs to be updated to a global value indicating the stride,  it is used in many places which is why this was easy to miss.

zontreck (approved):

Issue #462, opened 2020-12-31
OC8RC2 - AnimLock in dialog does not toggle when clicked


felkami:
**What version of OpenCollar are you using?**
OC8RC2

**What behavior did you expect?**
AnimLock to toggle when clicked

**What behavior did you see instead?**
When clicking AnimLock, option does not toggle

**What steps does someone need to take to reproduce the problem?**
Go to Animations, click AnimLock
Verify not showing checked
Click BACK, then click Animations
Verify still not checked

felkami:
Working

Issue #460, opened 2020-12-31
OC8RC2 - Upon running away, relay NOTIFY indicates activation even when relay is set to OFF


felkami:
**What version of OpenCollar are you using?**
OC8RC2

**What behavior did you expect?**
More neutral text that doesn't imply relay is being turned on when it was previously off, such as:
[21:48] Cara's Collar: Relay temporarily suppressed for 30 seconds due to safeword or clear all.
[21:49] Cara's Collar: Relay settings have been restored.

**What behavior did you see instead?**
[21:48] Cara's Collar: Relay temporarily disabled due to safeword or clear all. The relay will reactivate in 30 seconds
[21:49] Cara's Collar: Relay has been reactivated

**What steps does someone need to take to reproduce the problem?**
Use the runaway command

felkami:
working

Issue #457, opened 2020-12-31
OC8RC2 - CMD_ZERO passed when any LM Call with a '0' in it is opted in


felkami:
**What version of OpenCollar are you using?**
OC 8.8 RC2

**What behavior did you expect?**
When opting into NOTIFY, CMD_ZERO should not be opted into as well

**What behavior did you see instead?**
When iNum is checked against Nums, the '0' in the string '1002' is being seeing as optin for CMD_ZERO

This packet is received:
{"addon_name":"OpenCollar","pkt_type":"from_collar","iNum":0,"sMsg":"menu","kID":"<Removed for privacy>"}

**What steps does someone need to take to reproduce the problem?**
Run attached code and verify CMD_ZERO is passed. Comment out the update block, verify CMD_ZERO is not passed.

[Addon test.txt](https://github.com/OpenCollarTeam/OpenCollar/files/5756532/Addon.test.txt)

felkami:
Working on

Issue #456, opened 2020-12-30
AO Link is broken


zontreck:
The script is broken we need someone to test what AO versions are broken so this script can be fixed.
(Script is .AOLink.lsl in sr/c/spares)

felkami:
working on

PR #465, opened 2021-01-01
Fixes for several bugs

felkami:
oc_api
    fix: #457 - Switched to searching by list instead of substring
        Requires spacer ~ ie "1002~0~9001"
oc_anim
    fix: #456 - added LMChannel broadcast to suppress compatible AOs            
        Note: oc_couples doesn't need fix, LM-Boot code already there 
        Added check upon getting LM_SETTING_SAVE so that if saved setting is current anim, we don't stop it unnecessarily
    fix: #462          
oc_relay
    fix: #460


zontreck (approved):

zontreck:
Thank you for the contribution

Issue #461, opened 2020-12-31
OC8RC2 - Several dialog buttons lack respring or notify


felkami:
**What version of OpenCollar are you using?**
OC8RC2

**What behavior did you expect?**
Button presses to generate respring, or a NOTIFY stating %NO ACCESS%

**What behavior did you see instead?**
auth-public
    animations - nothing
    RLV>Force SIt>Strict Sit - no notify
    RLV>Relay - nothing
    Apps>Bookmarks - nothing
    Apps>Outifts - nothing
    Apps>Undress - nothing
    Apps>Titler - nothing
    Apps>Label - nothing
    Apps>Detach - nothing
    Leash>Anchor does not respring

auth-wearer
    RLV>RLV ON does not respring
    Animations>Pose>STOP does not respring
    Animations>Couples>{animation} does not respring

**What steps does someone need to take to reproduce the problem?**
With the appropriate auth level, click the stated button

PR #466, opened 2021-01-01
Fixes for 461 and other changes

felkami:
oc_anim
    Partial: #461  Fixed respring for poses>STOP
               Fixed respring and error notify for failed Public Animation access

oc_rlvsys
    Partial: #461  Fixed RLV ON respring
    
oc_rlvextension
    Partial: #461  Fixed Strict Sit no notify on respring
    
oc_relay
    Partial: #461  Fixed respring and error notify for public relay access
    
oc_bookmarks
    Partial: #461  Fixed respring and error notify for public relay access         

oc_outfits
    Made menu call case insensitive
    Partial: #461  Fixed respring and error notify for public relay access

### DIFF
--- a/src/Apps/oc_outfits.lsl
+++ b/src/Apps/oc_outfits.lsl
@@ -1,4 +1,5 @@
 /*
+oc_outfits
 This file is a part of OpenCollar.
 Copyright ©2020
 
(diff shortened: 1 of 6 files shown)


src/Apps/oc_outfits.lsl
```diff
@@ -218,7 +247,7 @@ UserCommand(integer iNum, string sStr, key kID) {
             RmCorelock();
             llSleep(1);
         }
-        if(sChangetype == "wear"){
+        if(sChangetype == "wear"){ //This looks like dead code TODO: Verify for removal?
```
zontreck, line 250: nope not dead code, the block above it performs the removal actions if applicable, then this block will perform the add operation

zontreck (approved):
Thank you for the contribution

PR #493, opened 2021-03-07
Faster and more reliable updater

zontreck:
This updater allows for multi-object owner updates and vastly improves 
the update time. There are minor protocol changes that require all staff 
approval before merging into master branch. If this will be merged, it 
must be tested a bit more thoroughly.


zontreck:
About ExperimentUpdater:  If you are on this experiment version, avatars other than the collar wearer can rez the updater and thus the wearer can update from it.  The process to do this remains the same, as it is designed to be a nearly seamless experience. This should not be considered a release version. Nor a full development snapshot, this is a experiment to see if this updater works and if we would like to replace the current updater using this new one.

What you will notice: You will notice that the updates occur much faster and more reliably.  Secondly, when you are using the new feature, you will see a black box rezzed at your avatar center, and you will then be leashed to it until the conclusion of the update.

Limitations: You must be within 5 meters of an updater you do not own in order to trigger the update. Can't do anything about the leash, it's just a neat little side effect for the time being of a mildly improper implementation of a rezzable leash post it would appear.

nirea:
So, faster sounds good.  But having one person update another person's collar seems like it would add a lot of complexity and security risk.

Is there a version that could just do the "faster" part?  And we talk about the other-person-update part separately?

zontreck:
yes. That piece can be taken out easily.

zontreck:
i'll make a commit that comments out the lines that access this feature.

zontreck:
Renamed the PR to be more accurate.  Just for the sake of my sanity and not wanting to completely redo the code if we end up adding the relay support, i haven't deleted the script for that, only made it inaccessible for the time being. If we opt not to go that route, i can completely remove references to that code with ease,.

silentwar-resident:
On this.. How is it possible because in do both has to be owned by the user not someone else. Did linden labs added new lsl feature lately?

silentwar-resident:
Opps sl not do. .... Grr phone autocorrect lol

zontreck:
I can't explain, i am sorry. Nirea deemed it to be a security risk. Whether it is or not, and i personally feel it isn't.... I just can't explain it on here.

silentwar-resident:
Ok no problem.

Sent from my iPhone

PR #501, opened 2021-03-14
meta(gha/lint): Add linting via lslint to GHA

lyseah:
This PR adds automatic linting of Pull Requests and Pushes (any branch).

run_lslint.sh is derivated from the original lslint test.sh script (see https://github.com/Makopo/lslint/blob/master/test.sh)

Action has been tested on ubuntu-latest, as well as debian buster (unavailable with GHA shared runners)

See example run here: https://github.com/lyseah/OpenCollar/runs/2108228764

### DIFF
--- a/.github/workflows/lslint.yml
+++ b/.github/workflows/lslint.yml
@@ -1,0 +1,25 @@
+name: LSLint
+
+on: [push, pull_request,workflow_dispatch]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    env:
+      LSLINT_VER: v1.2.6
+      LSLINT_ENV: linux
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: "Download prerequisite: Makopo/lslint"
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/lslint.yml
```diff
@@ -0,0 +1,25 @@
+name: LSLint
+
+on: [push, pull_request,workflow_dispatch]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    env:
+      LSLINT_VER: v1.2.6
```
zontreck, line 10: This could be made more efficient if we were to run the makefile to compile LSLint. No version number would be needed.  GHA is designed to compile software, Therefore it has a entire toolchain for compiling C and C++.

zontreck (approved):

.github/workflows/lslint.yml
lyseah: That is true, but please note that LSLINT has not received any updates since August 2020, so building lslint every time the workflow runs is a bit exessive imo.

PR #502, opened 2021-03-14
Remove old CircleCI config

zontreck:
This PR does a few things

* Remove Circle CI config
* Update github action config to only scan the src folder.
* Update github action config to not use wget, and instead compiles a fresh copy of lslint.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,0 @@
-version: 2
-jobs:
-  build:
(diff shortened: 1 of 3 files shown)


lyseah (approved):
Although I don't see the point in compiling the linter on each workflow run (as its last update was in August 2020), the changes run fine on my test environment (debian:buster) and report linting errors as expected. :)

Issue #506, opened 2021-03-17
Adding RLV folders with starting with same name fail silently


nescolet:
**What version of OpenCollar are you using?**
8.0.5
What behavior did you expect?
I expected this command "/1 {mypreffix}&slinkm" to attach item in slinkm subfolder inside #RLV folder, or at least being asked to choose between different RLV folders having that same string in name, as it was usual in collar versions before 8.

**What behavior did you see instead?**
I got nothing attached and since 8.0.5 is completely silent when attaching/removing RLV items via commands I had no clue if there was an error or not.
I was adviced by a former OpenCollar developer to rename "slinkm" folder to something like "slinkmhud", which I did and that worked around this issue, because the collar did not had to and fail to ask owner about which folder to choose.

**What steps does someone need to take to reproduce the problem?**
Create two subfolders in #RLV folder: one named "slinkm" other named "slinkmfeet", without quotes.
Type this command in chat: "/1 {mypreffix}&slinkm" replacing {mypreffix} with your own collar prefix

SilkieSabra:
Is this still an issue in 8.2.3?

nescolet:
> Is this still an issue in 8.2.3?

Yes, it is.

SilkieSabra:
Is this worth fixing? The obvious fix is to change folder names so they are not identical.

nescolet:
This can be useful even when for one wear it add command we don't type the the folder full name, then if opencollar finds two folders starting with that string it should display a menu with the numbered list and we can choose such one we want by the list number.

PR #508, opened 2021-03-19
Fix addons with public access, when in range, and provide a prompt when multiple folders are found with @findfolders

zontreck:
This PR fixes two issues:

1) Public access in oc_addons and oc_api
2) FindFolder in oc_folders has been updated to findfolders and now will show a menu prompt if there are multiple results.

### DIFF
--- a/src/collar/oc_api.lsl
+++ b/src/collar/oc_api.lsl
@@ -110,7 +110,7 @@ key g_kMenuUser;
 integer CalcAuth(key kID, integer iVerbose){
     string sID = (string)kID;
     // First check
-    if(llGetListLength(g_lOwner) == 0 && kID==g_kWearer && llListFindList(g_lBlock,[sID])==-1)
+    if(llGetListLength(g_lOwner) == 0 && kID==g_kWearer)
         return CMD_OWNER;
     else{
         if(llListFindList(g_lBlock,[sID])!=-1)return CMD_BLOCKED;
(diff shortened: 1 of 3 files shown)


zontreck:
Fixes: #506

src/collar/oc_api.lsl
```diff
@@ -110,7 +110,7 @@ key g_kMenuUser;
 integer CalcAuth(key kID, integer iVerbose){
     string sID = (string)kID;
     // First check
-    if(llGetListLength(g_lOwner) == 0 && kID==g_kWearer && llListFindList(g_lBlock,[sID])==-1)
```
nirea, line 113: Why would we not want to check the block list here?

src/collar/oc_folders.lsl
```diff
@@ -383,22 +410,23 @@ state active
...
+                for(x=0;x<end;x++)
+                {
+                    sAppend += "\n"+llList2String(lParam,x);
                 }
+                Dialog(kID, "There were multiple results found, please select one.\nOptions: \n\n"+sAppend, lParam, [], 0, g_iMenuUser, "FolderSearch");
+            } else if(llGetListLength(lParam) == 1){
+                ApplyFolderActions(sMsg);
+            } else if(llGetListLength(lParam) == 0)
+            {
+                llMessageLinked(LINK_SET, NOTIFY, "0No results were found", g_kMenuUser);
```
nirea, line 429: Is "0No" a typo or should it be "No"?

nirea (changes requested):

src/collar/oc_api.lsl
zontreck: Because that is a separate issue.  This only addresses the range issue, per you, the pull only addresses what it says it does. 

The issue of wearer being able to be blocked, and being able to be trusted, needs to be handled differently.

src/collar/oc_folders.lsl
lyseah: The 0/1 in front of the text determines if the wearer will be informed as well. This type of notification switching was implemented in 2015 or even earlier...

PR #512, opened 2021-03-24
Remove unused variables

RomkaSwallowtail:
removing meaningless and unused variables 'iSender' and 'iNum' from the
'g_lSensorDetails' list

### DIFF
--- a/src/collar/oc_dialog.lsl
+++ b/src/collar/oc_dialog.lsl
@@ -42,13 +42,12 @@ string BLANK = " ";
 integer g_iTimeOut = 300;
 integer g_iReapeat = 5;//how often the timer will go off, in seconds
 
-list g_lMenus;//12-strided list in form listenChan, dialogid, listener, starttime, recipient, prompt, list buttons, utility buttons, currentpage, button digits, auth level, sorted
+list g_lMenus;//13-strided list in form listenChan, dialogid, listener, starttime, recipient, prompt, list buttons, utility buttons, currentpage, button digits, auth level,extra info, sorted
 //where "list buttons" means the big list of choices presented to the user
 //and "page buttons" means utility buttons that will appear on every page, such as one saying "go up one level"
 //and "currentpage" is an integer meaning which page of the menu the user is currently viewing
-integer g_iStrideLength = 12;
+integer g_iStrideLength = 13;
 integer SENSORDIALOG = -9003;
-integer LOCALDIALOG = -9004;
 
 string NameURI(key kID){
     return "secondlife:///app/agent/"+(string)kID+"/about";
@@ -67,8 +66,6 @@ list g_lOwners;
 list g_lSensorDetails;
 integer g_bSensorLock;
 integer g_iSensorTimeout;
-integer g_iSelectAviMenu; //added to show URIs in menus june 2015 Otto(garvin.twine)
-integer g_iColorMenu;
 
 string SubstituteVars(string sMsg) {
         if (~llSubStringIndex(sMsg, "%NOACCESS"))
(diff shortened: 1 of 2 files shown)


src/collar/oc_dialog.lsl
```diff
@@ -42,12 +42,14 @@ string BLANK = " ";
 integer g_iTimeOut = 300;
 integer g_iReapeat = 5;//how often the timer will go off, in seconds
 
-list g_lMenus;//13-strided list in form listenChan, dialogid, listener, starttime, recipient, prompt, list buttons, utility buttons, currentpage, button digits, auth level,extra info,  sorted
+list g_lMenus;//12-strided list in form listenChan, dialogid, listener, starttime, recipient, prompt, list buttons, utility buttons, currentpage, button digits, auth level, sorted
```
nirea, line 45: I'm trying to think of what this extrainfo piece of data would be for.

I remember once upon a time discussing that instead of plugins having to keep track of menu UUIDs, it would be nice if they could just pass in their own arbitrary names or modes that then they would get back when the DIALOG_RESPONSE came.  That's my best guess as to why that was ever here, but yeah it doesn't look like it was being used that way.

You have any more history on this?

nirea: Only other comment is that this looks like the kind of change where it would be really easy to miss fixing a llList2String call somewhere.  I don't have time to check for all of those right now but it seems like someone should.  Or else have some rigorous testing on it inworld.

RomkaSwallowtail: perhaps 'extraInfo' was supposed to be used in the future, but for a long time it has not been used at all
I got the idea to use 'extraInfo' instead of the 'g_iSelectAviMenu' and 'g_iColorMenu' global variables to select the dialog mode
do not merge yet, I will make a new pull request

RomkaSwallowtail: now 'g_iSelectAviMenu' and 'g_iColorMenu'  global variables are not used to change the dialog mode,
used 'extrainfo' instead

RomkaSwallowtail: p.s. i tested in the world it works

src/collar/oc_particle.lsl
```diff
@@ -155,7 +155,7 @@ FindLinkedPrims() {
 
     for (i=-1; i<=linkcount;++i) {
         string sPrimName = llToLower(llStringTrim(llList2String(llGetLinkPrimitiveParams(i,[PRIM_NAME]),0),STRING_TRIM));
-        if (llToLower(sPrimName) == "leashpoint" || llToLower(sPrimName) == "ooc") {
```
nirea, line 158: Are you sure we don't need the llToLower?  I have a vague memory of code like this being used to account for uppercased menu buttons.

src/collar/oc_dialog.lsl
```diff
@@ -597,28 +589,24 @@ state active
         string sButtons=llDumpList2String(lButtons,"`");
         lParams=llListReplaceList(lParams,[sButtons],3,3);
         //fake fresh dialog call with our new buttons in place, using the rest of the information we were sent
-        llMessageLinked(LINK_SET,DIALOG,llDumpList2String(lParams,"|"),(key)llList2String(lSensorInfo,3));
+        llMessageLinked(LINK_THIS,DIALOG,llDumpList2String(lParams,"|"),kID);
```
nirea, line 592: 👍

src/collar/oc_dialog.lsl
```diff
@@ -161,13 +160,13 @@ list g_lColors = [
...
 
-Dialog(key kRecipient, string sPrompt, list lMenuItems, list lUtilityButtons, integer iPage, key kID, integer iWithNums, integer iAuth,string extraInfo, integer iSorted)
+Dialog(key kRecipient, string sPrompt, list lMenuItems, list lUtilityButtons, integer iPage, key kID, integer iWithNums, integer iAuth, string extraInfo, integer iSorted)
 {
     //calculate page start and end
     integer iNumitems = llGetListLength(lMenuItems);
     integer iStart = 0;
     integer iMyPageSize = g_iPagesize - llGetListLength(lUtilityButtons);
-    if (g_iSelectAviMenu) { //we have to reduce buttons due to text length limitations we reach with URI
+    if (extraInfo == "avimenu") { //we have to reduce buttons due to text length limitations we reach with URI
```
nirea, line 169: If you want a menu mode, why not just add an argument for that to the Dialog function?  That would seem to be the clearest way to express what's going on.  The name "extrainfo" seems harder for people to understand how it's used.  Even just changing the name from "extrainfo" to "menuMode" would be OK I think.

src/collar/oc_particle.lsl
RomkaSwallowtail: take a look at line 157 - there is already    sPrimName = llToLower(

src/collar/oc_dialog.lsl
RomkaSwallowtail: I suppose in the future extrainfo can be used not only to change dialog mode

RomkaSwallowtail: particles - #530
dialog - #531

PR #514, opened 2021-03-26
Update oc_update_seed.lsl

SilkieSabra:
Updating the instructions for the update_seed


nirea (changes requested):
Can you re-target this pull request to the features-2021-q2 branch instead of master?

PR #530, opened 2021-04-01
Update oc_particle.lsl - removе double

RomkaSwallowtail:
llToLower (sPrimName) on line 158 does not make sense,
because line 157 already contains sPrimName = llToLower (llStringTrim (...


Medea-Destiny:
Simple efficiency improvement, no need to do an llToLower comparison on a string that has already been through llToLower

PR #535, opened 2021-04-04
Fix problem with addon optin

devy-null:
It was not using the list it modified in some examples
Now using the correct code from unwelder instead


nirea:
Hi @devy-null .  Thank you for the patch!

I'm trying to decide if this should get merged and released with the features-2021-q2 branch or if we need a bugfix release sooner than that.  How broken is the user experience without this fix?

I'm going to edit the PR and target it at that feature branch for now.

PR #540, opened 2021-04-12
don't forget set g_iMenuStride !

RomkaSwallowtail:
don't forget to set constant g_iMenuStride=3
this is important for correct handling of g_lMenuIDs list

I found that this was not done in the 8.0.5000 release! it can cause bugs!

### DIFF
--- a/src/collar/oc_titler.lsl
+++ b/src/collar/oc_titler.lsl
@@ -193,7 +193,7 @@ integer g_iOffset=8;
 vector g_vColor = <1,1,1>;
 string g_sTitle;
 list g_lMenuIDs;
-integer g_iMenuStride;
+integer g_iMenuStride=3;
 list g_lOwner;
 list g_lTrust;
 list g_lBlock;
--- a/src/collar/oc_titler.lsl
+++ b/src/collar/oc_titler.lsl
@@ -366,7 +366,7 @@ state active
             }
         }else if (iNum == DIALOG_TIMEOUT) {
             integer iMenuIndex = llListFindList(g_lMenuIDs, [kID]);
-            g_lMenuIDs = llDeleteSubList(g_lMenuIDs, iMenuIndex - 1, iMenuIndex +3);  //remove stride from g_lMenuIDs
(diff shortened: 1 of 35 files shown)


src/collar/oc_titler.lsl
```diff
@@ -193,7 +193,7 @@ integer g_iOffset=8;
 vector g_vColor = <1,1,1>;
 string g_sTitle;
 list g_lMenuIDs;
-integer g_iMenuStride;
+integer g_iMenuStride=3;
```
nirea, line 196: Line 297 of this script looks like it might be doing the wrong thing now that this value has changed.  Can you check that?

nirea: And double check the usage in the other scripts too?

RomkaSwallowtail: checked twice, three times or more times all scripts 

look at line 89 - g_iMenuStride sets the length of the replaced part of g_lMenuIDs list 
this part contains **3** list items: [kID, kMenuID, sName] 
on line 90 these **3** elements are added to g_lMenuIDs list : [kID, kMenuID, sName] 

on line 297 g_iMenuStride sets the length of the removed part of g_lMenuIDs list

if g_iMenuStride value is not set, that is, it is equal to 0, then in lines 89 and 297, the wrong number of wrong elements of g_lMenuIDs list will be replaced and deleted 

I don’t understand why you forgot that in previous versions of collar scripts this parameter g_iMenuStride was always set as constant.
usually it was equal to 3, sometimes more, as for example in the oc_dialog script

RomkaSwallowtail: you can get rid of g_iMenuStride altogether and replace it with a digit
in fact, g_iMenuStride is almost everywhere 3 
for example, use "iMenuIndex + 1" instead of "iMenuIndex - 2 + g_iMenuStride" the same as "iMenuIndex - 2 + 3"

RomkaSwallowtail:
I guess I'm trying in vain ...

RomkaSwallowtail:
nevermind

Issue #534, opened 2021-04-04
Public chat is channel 0


RomkaSwallowtail:
the manual on the site https://opencollar.cc/docs/Prefixes,-Chat-Commands-and-Other-Non-Menu-Functions says: 

Public chat is channel 0. You can turn this off so the collar doesn’t listen to chat commands in local chat.

    [prefix] channel 0 : enables the collar to listen to commands in local chat.
    [prefix] channel -1 : disables the collar from listening to commands in local chat.

**but in reality this function does not work! collar always listen local chat.**

zontreck:
This is intentional.  Not a bug

Apr 3, 2021 11:51:28 PM Romka ***@***.***>:

RomkaSwallowtail:
in this case, you need to correct the manual so as not to mislead users

zontreck:
Let the team know.  I'm no longer in the team just telling you that it's been this way since 3.x

Apr 4, 2021 1:59:38 AM Romka ***@***.***>:

SilkieSabra:
Since I write the manual I can test this.  However if it doesn't work there should be a fix because also since 3.x there has to be a way to turn off the collar listening to local chat.

RomkaSwallowtail:
@SilkieSabra it doesn't work, you can make sure.
if you like, I have ideas how to fix this without increasing the script code too much

SilkieSabra:
OpenCollar V8.1 uses the negative channel to close channel 0

PR #541, opened 2021-04-13
listening to local chat 

RomkaSwallowtail:
enable / disable listening to local chat  #534

i suggest using a negative value of the g_iChannel variable to disable listening to local chat

### DIFF
--- a/src/collar/oc_api.lsl
+++ b/src/collar/oc_api.lsl
@@ -178,7 +178,7 @@ integer g_iChatListener;
 DoListeners(){
     if (g_iListener) llListenRemove(g_iListener);
     if (g_iChatListener) llListenRemove(g_iChatListener);
-    g_iListener = llListen(llAbs(g_iChannel), "","","");
+    g_iListener = llListen(g_iChannel, "","","");
     if (g_iChannel > 0) g_iChatListener = llListen(0,"","","");
 }
 integer g_iRunaway=TRUE;
@@ -330,13 +330,7 @@ UserCommand(integer iAuth, string sCmd, key kID){
         string sCmdx = llToLower(llList2String(lCmd,0));
                 
         if(sCmdx == "channel"){
-            integer chan = (integer)llList2String(lCmd,1);
-            if (chan == -1) g_iChannel = -llAbs(g_iChannel);
-            else if (chan == 0) g_iChannel = llAbs(g_iChannel);
-            else if (chan > 0) {
-                if (g_iChannel < 0) g_iChannel = -chan;
-                else g_iChannel = chan;
-            }
+            g_iChannel = (integer)llList2String(lCmd,1);
             llMessageLinked(LINK_SET, LM_SETTING_SAVE, "global_channel="+(string)g_iChannel, kID);
         
         } else if(sCmdx == "prefix"){
--- a/src/collar/oc_api.lsl
+++ b/src/collar/oc_api.lsl
@@ -333,7 +333,10 @@ UserCommand(integer iAuth, string sCmd, key kID){
             integer chan = (integer)llList2String(lCmd,1);
             if (chan == -1) g_iChannel = -llAbs(g_iChannel);
             else if (chan == 0) g_iChannel = llAbs(g_iChannel);
-            else if (chan > 0) g_iChannel = chan;
+            else if (chan > 0) {
+                if (g_iChannel < 0) g_iChannel = -chan;
+                else g_iChannel = chan;
+            }
             llMessageLinked(LINK_SET, LM_SETTING_SAVE, "global_channel="+(string)g_iChannel, kID);
         
         } else if(sCmdx == "prefix"){
--- a/src/collar/oc_api.lsl
+++ b/src/collar/oc_api.lsl
@@ -173,16 +173,13 @@ PrintAccess(key kID){
     //llSay(0, sFinal);
 }
 
-list g_lActiveListeners;
+integer g_iListener;
+integer g_iChatListener;
 DoListeners(){
-    integer i=0;
-    integer end = llGetListLength(g_lActiveListeners);
-    for(i=0;i<end;i++){
-        llListenRemove(llList2Integer(g_lActiveListeners, i));
-    }
-    
-    g_lActiveListeners = [llListen(g_iChannel, "","",""), llListen(0,"","",""),  llListen(g_iInterfaceChannel, "", "", "")];
-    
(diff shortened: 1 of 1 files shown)


src/collar/oc_api.lsl
```diff
@@ -173,16 +173,13 @@ PrintAccess(key kID){
     //llSay(0, sFinal);
 }
 
-list g_lActiveListeners;
```
RomkaSwallowtail, line 176: it makes no sense to create a list for only two variables

src/collar/oc_api.lsl
```diff
@@ -173,16 +173,13 @@ PrintAccess(key kID){
...
-    integer end = llGetListLength(g_lActiveListeners);
-    for(i=0;i<end;i++){
-        llListenRemove(llList2Integer(g_lActiveListeners, i));
-    }
-    
-    g_lActiveListeners = [llListen(g_iChannel, "","",""), llListen(0,"","",""),  llListen(g_iInterfaceChannel, "", "", "")];
-    
+    if (g_iListener) llListenRemove(g_iListener);
+    if (g_iChatListener) llListenRemove(g_iChatListener);
+    g_iListener = llListen(llAbs(g_iChannel), "","","");
```
RomkaSwallowtail, line 181: to listen to a private channel, a positive g_iChannel value is always used

src/collar/oc_api.lsl
```diff
@@ -173,16 +173,13 @@ PrintAccess(key kID){
...
+    if (g_iChannel > 0) g_iChatListener = llListen(0,"","","");
```
RomkaSwallowtail, line 182: negative g_iChannel means not listening to local chat

src/collar/oc_api.lsl
```diff
@@ -470,6 +470,7 @@ state active
             g_iInterfaceChannel = (integer)("0x" + llGetSubString(g_kWearer,30,-1));
             if (g_iInterfaceChannel > 0) g_iInterfaceChannel = -g_iInterfaceChannel;
         }
+        llListen(g_iInterfaceChannel, "", "", "");
```
RomkaSwallowtail, line 473: this channel should always be listened to

src/collar/oc_api.lsl
```diff
@@ -333,7 +330,10 @@ UserCommand(integer iAuth, string sCmd, key kID){
         string sCmdx = llToLower(llList2String(lCmd,0));
                 
         if(sCmdx == "channel"){
-            g_iChannel = (integer)llList2String(lCmd,1);
+            integer chan = (integer)llList2String(lCmd,1);
```
RomkaSwallowtail, line 333: [prefix] channel -1 : disables the collar from listening to commands in local chat,
[prefix] channel 0 : enables the collar to listen to commands in local chat,
[prefix] channel [new number] : change the channel number.

src/collar/oc_api.lsl
```diff
@@ -333,7 +330,10 @@ UserCommand(integer iAuth, string sCmd, key kID){
...
+            if (chan == -1) g_iChannel = -llAbs(g_iChannel);
+            else if (chan == 0) g_iChannel = llAbs(g_iChannel);
+            else if (chan > 0) g_iChannel = chan;
```
RomkaSwallowtail, line 336: [prefix] channel -1 : disables the collar from listening to commands in local chat,
[prefix] channel 0 : enables the collar to listen to commands in local chat,
[prefix] channel [new number] : change the channel number.

src/collar/oc_api.lsl
```diff
@@ -333,7 +330,13 @@ UserCommand(integer iAuth, string sCmd, key kID){
...
                 
         if(sCmdx == "channel"){
-            g_iChannel = (integer)llList2String(lCmd,1);
+            integer chan = (integer)llList2String(lCmd,1);
+            if (chan == -1) g_iChannel = -llAbs(g_iChannel);
+            else if (chan == 0) g_iChannel = llAbs(g_iChannel);
+            else if (chan > 0) {
+                if (g_iChannel < 0) g_iChannel = -chan;
+                else g_iChannel = chan;
+            }
```
RomkaSwallowtail, line 339: [prefix] channel -1 : disables the collar from listening to commands in local chat,
[prefix] channel 0 : enables the collar to listen to commands in local chat,
[prefix] channel [new number] : change the channel number.

src/collar/oc_api.lsl
nirea: If someone tried to set a negative one, then silently switching that to positive seems like it would be confusing.

src/collar/oc_api.lsl
nirea: Not sure the logic here is correct.  Why inverting the sign with `-chan`?

src/collar/oc_api.lsl
RomkaSwallowtail: ok, I agree, this method is not suitable, I will look for another

RomkaSwallowtail:
ok, I agree, this method is not suitable, I will look for another

but what about which is described on the site the command "[prefix] channel -1" for disables the collar from listening to commands in local chat?
now this command just sets listening to the private channel -1
suggest a way ...

src/collar/oc_api.lsl
RomkaSwallowtail: ok, let's leave the negative channel unchanged

src/collar/oc_api.lsl
```diff
@@ -173,16 +173,13 @@ PrintAccess(key kID){
...
-    for(i=0;i<end;i++){
-        llListenRemove(llList2Integer(g_lActiveListeners, i));
-    }
-    
-    g_lActiveListeners = [llListen(g_iChannel, "","",""), llListen(0,"","",""),  llListen(g_iInterfaceChannel, "", "", "")];
-    
+    if (g_iListener) llListenRemove(g_iListener);
+    if (g_iChatListener) llListenRemove(g_iChatListener);
+    g_iListener = llListen(g_iChannel, "","","");
+    if (g_iChannel > 0) g_iChatListener = llListen(0,"","","");
```
RomkaSwallowtail, line 182: negative g_iChannel means not listening to local chat

src/collar/oc_api.lsl
RomkaSwallowtail: ok, let's leave the negative channel unchanged

RomkaSwallowtail:
new option: I suggest using negative private channel setting to disable listening to local chat channel

PR #543, opened 2021-04-20
listening to local chat (another way)

RomkaSwallowtail:
i suggest using a separate command to enable/disable listening to local chat,
for example: "channel on" (or "channel 0") and "channel off", or what do you suggest.
this will avoid conflict with the "channel [new number]" command


RomkaSwallowtail:
nevermind

Issue #556, opened 2021-04-27
Trusted can unlock collar that was locked by owner


1Acadia:
Version 8.0.5

I expected my collar, when locked by my owner, to not be unlocked by a Trusted person
I had my owner lock it, and my alt who is on the Trusted list was able to unlock it, but then unable to re-lock.  I expected my owner's permissions to be greater than a trusted and trusted not able to unlock.
To re-create the issue, have your collar locked by your owner and then have someone on the trusted list try to unlock.  We verifed the issue with another person and the same thing happened.

mistressohm:
Is this not Working As Intended?  Perhaps I'm thinking of a different scenario, but Trusted persons are kind of like backup Owners, able to do some things but not all.    

If your collar is locked and you have need to unlock it (and cannot unlock it yourself), you need your Owner.  But, if there is someone Trusted on the collar, they can help get you out of whatever predicament you're in... however, only your Owner can lock your collar again once this happens.  This is to keep subs from 'cheating' on their collar - their Owner will find out that the collar was unlocked ere long.   

However, there is also a possibility that the collar is like a safe - if the Owner has locked it, only the Owner can unlock.  If a Trusted locks your collar, then any Trusted can unlock, along with the Owner.  In this case, yes, the permissions are not correct.

I am intentionally staying away from Safeword/Runaway scenarios - the one is more for undoing RLV restrictions and getting back to normal without unlocking the collar, and the other is for getting out of a bad D/s situation permanently, erasing all ownership and reverting the collar to 'out of the box'.

SilkieSabra:
This is not how the collar should work, Trusted has never been able to unlock a collar if an owner has locked it.

Pingout:
This issue can be closed. The fix has been included in the next quarterly release. Thank you.

1Acadia:
Thank you, Ping

PR #557, opened 2021-04-30
Update oc_core.lsl

Pingout:
Removed ability for Trusted to unlock the collar

### DIFF
--- a/src/collar/oc_core.lsl
+++ b/src/collar/oc_core.lsl
@@ -322,7 +322,7 @@ UserCommand(integer iNum, string sStr, key kID) {
             llMessageLinked(LINK_SET, LM_SETTING_SAVE, "global_locked="+(string)g_iLocked,"");
             llMessageLinked(LINK_SET, NOTIFY, "1%WEARERNAME%'s collar has been locked", kID);
             llPlaySound(g_sLockSound,1);
-        } else if(llToLower(sChangetype) == "unlock" && (iNum == CMD_OWNER || iNum == CMD_TRUSTED) && !g_iWelded){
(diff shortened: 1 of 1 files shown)


SilkieSabra (approved):
This may have been added recently, I don't recall any collar prior to V8 having Trusted able to unlock.

Pingout:
Reference Issue #556

src/collar/oc_core.lsl
```diff
@@ -322,7 +322,7 @@ UserCommand(integer iNum, string sStr, key kID) {
             llMessageLinked(LINK_SET, LM_SETTING_SAVE, "global_locked="+(string)g_iLocked,"");
             llMessageLinked(LINK_SET, NOTIFY, "1%WEARERNAME%'s collar has been locked", kID);
             llPlaySound(g_sLockSound,1);
-        } else if(llToLower(sChangetype) == "unlock" && (iNum == CMD_OWNER || iNum == CMD_TRUSTED) && !g_iWelded){
+        } else if(llToLower(sChangetype) == "unlock" && iNum == CMD_OWNER && !g_iWelded){
```
nirea, line 325: Is trusted able to lock the collar?  We wouldn't want someone able to lock but not unlock, so you might need a corresponding change on that side.

Pingout: > Is trusted able to lock the collar? We wouldn't want someone able to lock but not unlock, so you might need a corresponding change on that side.

Trusted has never had access to lock the collar as far as I know. I think the change was an ill-advised response to someone getting trapped in furniture, hoping a Trusted could get them out of the trap when their Owner was off line. Bottom line is a Trusted should not be able to undo something an Owner set. And since the Trusted can not Lock the collar they should not be able to Unlock it.

PR #558, opened 2021-05-03
make ao compatible with addon system in v8

tayaphidoux:
because of the use of lists in this ao there was difficulty getting it to accept most of the necessary functions for addon support, also the script uses the timer function for stand timing, the addon system for the collar is aggressive and requires constant ping pong updates from the item for it to be displayed constantly int he menu, these changes and the addon script i made allow the ao to function along side the collar and always be accessible.

https://github.com/OpenCollarTeam/OpenCollar/pull/559 - addon script mentioned above


Omnia2021:
I have been using the AO and find it intuitive and responsive, no issues of note experienced in everyday usage. I also added my own animations and that went very smooth as well.

Omnia2021:
I have been using the AO and find it intuitive and responsive, no issues of note experienced in everyday usage. I also added my own animations and that went very smooth as well.

Pingout (approved):
Working good, thanks Taya!

Issue #538, opened 2021-04-06
Potential addons ``llResetScript()`` loop 


TechTheAwesome:
What version of OpenCollar are you using?
8.0.5000

What behavior did you expect?
``oc_addon_template`` to be answered with pong when ping is fired.

What behavior did you see instead?
Due to ``g_iLMLastRecv`` has the value of 0, thus always script reset every time ping is fired, triggering setting request "ALL" which can be laggy with many addons. 

What steps does someone need to take to reproduce the problem?
One solution is to set ``g_iLMLastRecv = llGetUnixTime();`` on ``state_entry()`` or on receiving an approved from the collar. 
``oc_cuff``doesn't have this issue due to ``g_iLMLastRecv`` being set on HTML version check received.
As long as ``llGetUnixTime() > (g_iLMLastRecv + (5 * 60))`` doesn't return ``true`` during the first ``timer()`` event it won't trigger a restart before addons can receive a ``pong`` answer from collar to update ``g_iLMLastRecv``.

SilkieSabra:
Thanks Tech, can you make the fix you suggest and do a pull request?

PR #561, opened 2021-05-05
on attach connect on removal disconnect

tayaphidoux:
make a clean connection and disconnection when worn or taken off, also implements the timer fix for the reset loop mentioned https://github.com/OpenCollarTeam/OpenCollar/issues/538


Pingout (approved):
This update worked well in Taya's oc_ao_addon.lsl script

Issue #524, opened 2021-03-29
Cuffs visual lock/unlock similar to Collar


Pingout:
What version of OpenCollar are you using?
OpenCollar 8.0.5 - OC Cuffs 8.0 Beta 
What behavior did you expect?
It would be nice to have a lock/unlock feature similar to the Collar where there are two locks, one open and one closed, which show and hide based on the locked/ unlocked status of the cuff
What behavior did you see instead?
This feature is not available in cuffs
What steps does someone need to take to reproduce the problem?

Pingout:
In my Github Issue #525, I'm asking for the visual lock/unlock of a lock object. Looking through the oc_cuffs script, the subroutine is already there to do what I'm asking. It's line 359 in the oc_cuffs script, starting with ToggleLock

But it doesn't work, and I can't see why

Pingout:
Fixed by Safra

PR #563, opened 2021-05-15
Update oc_cuff.lsl

Pingout:
Visible lock fix reference Issue #524 with thanks to Safra Nitely

This version also includes the chain gravity fix (Chains had too much sag)


nirea:
Thank you!

PR #567, opened 2021-05-25
Update oc_cuff.lsl

Pingout:
Here is a fully working oc_cuffs locking system thanks to Safra, in response to Issue #524 .
Our first attempt at a fix had a bug in which SyncLock was not working correctly. This script fixes that problem. The cuff locks will now display the ClosedLock cuffs when the collar is locked, and OpenLock when the collar is unlocked with SyncLock active


Pingout:
Included in Safra's fix

PR #568, opened 2021-05-25
Update .themes

Pingout:
This is an example [Name].theme notecard, from the Version 8.x OpenCollar Stock collar.


SilkieSabra (approved):
This replaces the old themes notation with the V8 themes for the stock collar, useful as examples for people wanting to create their own

PR #569, opened 2021-05-25
Delete .textures

Pingout:
The .textures card is no longer used in Version 8.x


SilkieSabra (commented):
This is old stuff ! let it slide into the tide...

SilkieSabra (approved):

PR #572, opened 2021-06-15
Update oc_api.lsl

Pingout:
This is Romka's fix that got deleted re: Issue #534 and Issue #541

### DIFF
--- a/src/collar/oc_api.lsl
+++ b/src/collar/oc_api.lsl
@@ -2,10 +2,7 @@
 /*
 This file is a part of OpenCollar.
 Copyright ©2021
-
-
 : Contributors :
-
 Aria (Tashia Redrose)
     *June 2020       -       Created oc_api
       * This implements some auth features, and acts as a API Bridge for addons and plugins
@@ -16,7 +13,6 @@ Felkami (Caraway Ohmai)
 et al.
 Licensed under the GPLv2. See LICENSE for full details.
 https://github.com/OpenCollarTeam/OpenCollar
-
 */
 list g_lOwner;
 list g_lTrust;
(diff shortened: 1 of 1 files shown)


Pingout:
Here's the expected behavior:
the Channel command allows the user to switch from the default hidden channel to any other hidden channel
so [prefix] channel 5 would make channel 5 your hidden channel
previously channel 0 would always be active, any command given in local chat will be followed
this script fix changes that
Now, when you assign a negative channel, ie. you type [prefix] channel -5
ONLY channel -5 will work, the open chat channel 0 is ignored
so typing [prefix] channel -5 will change the hidden channel to -5, but after that if you type [prefix] channel 1 that does nothing
to restore the default channel 0 and channel 1 you would need to type /-5 [prefix] channel 1

Trinkitz:
Tested this at some length today. Switching to a negative channel does prevent the collar from seeing local chat, and chat commands delivered to the negative channel number worked fine. However, when set to a negative channel, the collar does not see the safeword when typed in local; the safeword must be typed on the proper channel in order to work. This is in contrast to when the collar is set to a positive channel, where typing the safeword in local works regardless of what (positive) channel the collar is set to. This does open up the possibility for an owner to all but disable the safeword through a few RLV restrictions and setting the channel to a 4-digit negative number for instance. Hope this helped; wave if I can help any further.

Pingout:
It looks like the Channel command is a higher priority command and is only available to the owner of the collar. If the collar owner changes the channel, the wearer can not change it back, unless the wearer is an owner

Trinkitz:
Just an addendum to my earlier comment. It appears as though "runaway" behaves in much the same manner as the safeword. It will not work in local when the collar is set to a negative channel.

SilkieSabra:
Safeword and Runaway always have to be typed on a channel if channel 0 is turned off.  If we are going back to older expected behavior, that is part of it.  The only new thing is that you have to assign a negative channel to do that.

Trinkitz:
Yes. I think if this fix is implemented, we may consider adding something like "If you use a negative channel, remember that ALL local chat commands (including safeword and runaway) must be typed in the appropriate channel; they will no longer work in local." to the Guide.

SilkieSabra (approved):
Omnia and Trinkitz tested it as well as Ping.  I call that good to go.

PR #574, opened 2021-06-23
Update oc_core.lsl

Pingout:
Added a line to iAuth only Owner and Wearer to get the Access List

### DIFF
--- a/src/collar/oc_core.lsl
+++ b/src/collar/oc_core.lsl
@@ -2,20 +2,14 @@
 /*
 This file is a part of OpenCollar.
 Copyright ©2021
-
-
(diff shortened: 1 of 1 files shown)


Pingout (commented):
This worked for both Omnia and me,

Omnia2021 (approved):
Works as advertised fixes an access flaw that could have been exploited without the wearer's knowledge.

SilkieSabra (approved):
let's get this into 8.1

Trinkitz:
This worked for me as well; neither Trusted, Public nor Group access level users could view the Access List.

PR #583, opened 2021-06-29
Update to oc_cuff includes Animation Priority

Pingout:
Thanks to Safra!


Pingout:
Ref issue #584 final testing as we prepare the OpenCollar Cuffs for Release Candidate

Issue #581, opened 2021-06-28
When Folders > Configure is set Trusted access is off, Trusted gets Access Denied to entire collar


SilkieSabra:
We tested this with 8.0.5 and 8.1 RC3.

V 8.0.5
Imp Trollop, wearer, Amanda Darkheart, owner, Roan, Trusted.  When Owner Configured Folders to remove Trusted access (public was also removed but we did not test public access); A Trusted could not Browse RLV Folders and got "Access Denied to Browse #RLV" which is inaccurate since it's only Folders that should be Access Denied.  Then when the menu was refreshed once, Trusted began to get that message on opening the collar menu.  Collar Access still worked as expected.

V 8.1 RC3
Same as V 8.0.5 except now when Trusted opens the collar menu, the Folders menu also pops up and Access Denied to Folders is sent.  Trusted then cannot access Browse, as expected; but also cannot access any other submenu in the collar because the front menu disappears when touched.  This happened three times.  The fourth time, the wearer got the popup when Trusted opened the collar menu.  Did the whole round a second time, and Trusted got Access Denied when attempting to Browse Folders.  Pressed Back  from there and the Folders menu reappeared.  Pressed Ignore, went back to main menu, Wearer started getting Access Denied messages.

Trinkitz:
Using the fix script, and version 8.1 RC-3 collar, when I remove Trusted access to folders, Trusted sees this message: "Access Denied to browsing #RLV" when they click on the collar, and gets these two menus from that single click: https://gyazo.com/712800ada76161ea24c31ab3ed7cd68e

If I give Trusted access again, all is fine. Hope this helps.

Pingout:
> Using the fix script, and version 8.1 RC-3 collar, when I remove Trusted access to folders, Trusted sees this message: "Access Denied to browsing #RLV" when they click on the collar, and gets these two menus from that single click: https://gyazo.com/712800ada76161ea24c31ab3ed7cd68e
> 
> If I give Trusted access again, all is fine. Hope this helps.

Are you saying Trusted sees this message right away, or do they see it when they try to access the denied area?

Trinkitz:
I clarified in the previous comment. When Trusted access is removed, Trusted sees two menus and receives that message when they click the collar.

SilkieSabra:
[...]

Issue #570, opened 2021-06-02
Hide ~ folder names in Folder listings


deserted1:
What version of OpenCollar are you using?
8.0.5
What behavior did you expect?
The awesome that is.
What behavior did you see instead?
The awesome that is.
What steps does someone need to take to reproduce the problem?
Wear anything by dhb that requires an autohide then look in #RLV then list your #Folders using the collar.

I am asking for a way to hide ~ folders just as we do with . folders in the listings.  Maybe as an option? as it seems some creators like using ~ folders for their temporary names and wont allow you to move them into .folder names.

Thank you~  <- Get it

SilkieSabra:
a ~[folder] is rlv standard string for a folder that can be given directly to RLV.  I agree I wish there was a way to autohide these, otherwise to keep your RLV folders trimmed you need to delete them until the next time the trap comes up and gives them to you all over again.  If the folder goes into .folders, it can't be seen by the RLV search the trap uses and so you will end up with multiple folders.

deserted1:
> a ~[folder] is rlv standard string for a folder that can be given directly to RLV. I agree I wish there was a way to autohide these, otherwise to keep your RLV folders trimmed you need to delete them until the next time the trap comes up and gives them to you all over again. If the folder goes into .folders, it can't be seen by the RLV search the trap uses and so you will end up with multiple folders.

Silkie, I am not asking to move them into the . folders, I am asking if we can specify a way to filter then from the Folders menu inside the collar. There is no reason for an end user to see all those if they choose not to.  Why not parse the foldername and if it begins with a ~ dont show it in the llDialog?  (this can be an option if some choose to see them)

Medea-Destiny:
Optional filtering of folders starting with ~ in https://github.com/OpenCollarTeam/OpenCollar/tree/Medea-4.1-1

deserted1:
Thank you Madea!

PR #591, opened 2021-07-04
Update oc_folders.lsl

Medea-Destiny:
*Fix issue #570, Allow hiding folders starting with ~ via HideTilde option, defaults to ON. 
*Fix issue  #581, filtering input to UserCommand to only folder-auth check actual folder commands rather than folder-authing and processing EVERYTHING THE COLLAR DOES BY ANY USER


Pingout:
This appears to work as advertised: 
a new checkbox shows up in the RLV Folders Configure menu, seems to work good, and I have access to the checkbox as Omnia's owner, but she gets a message saying Access Denied to folders configuration

Issue #580, opened 2021-06-28
Print settings should be auth limited


Medea-Destiny:
Version 8.0-8.1rc3

Settings menu -> print to dump settings does not currently have an auth check. These settings shouldn't be exposed to everyone. We should have an if(iAuth==CMD_OWNER || iAuth==CMD_WEARER) here.

Medea-Destiny:
Fix in https://github.com/OpenCollarTeam/OpenCollar/tree/Medea-4.1-1/src/collar

Issue #579, opened 2021-06-28
Listener issues: Listening on local chat, and the interface channel


Medea-Destiny:
Version 8.1 RC3

References: #572 #534 #541

As per current release candidate version, there's a fix to allow switching off channel 0 by setting the collar's channel to a negative channel.

Chatting on negative channels may be problematic - at least as far as I know this is still a TPV-only capability, and the standard SL viewer does not allow it. Ideally people should be able to switch off the local chat listener while staying to a positive channel, particularly as people are very used to using channel 1. I see the best solution to this as being adding a button to the settings menu which toggles listening on channel 0 on and off.

I notice that at some point in implementing the current change, the interface channel listener has been lost. This is a problem as the interface channel is what is used for collars to check that a second collar doesn't get worn at the same time. the listener for the interface channel should be returned to the code, and the opencollar=yes chat on startup restored to allow this functionality - as it stands, multiple collars can be worn.

The code in the current oc_api script only uses the interface channel for this code. Historically the interface channel has allowed external devices to communicate with the collar. Items which trigger the collar menu - particularly useful when clicking the collar is difficult to do - use the menuto command. Other items may use the interface channel's capability to send commands to the collar as if chatting on the collar channel. Neither of these capabilities are in the current code.

I suggest as minimum reimplementing the menuto function. The requirement for passing arbitrary commands for auth via the interface channel is of more limited value; the add-ons feature should replace this kind of need. However it might be considered for backwards compatibility.

Medea-Destiny:
Fixes in https://github.com/OpenCollarTeam/OpenCollar/tree/Medea-4.1-1/src/collar

Issue #566, opened 2021-05-25
Setgroup not clearing when group is unchecked


SilkieSabra:
In V8.0.5, several users have reported that Group Access was checked, then unchecked, but @setgroup=0 was not cleared.  It may be intermittent as at least one person was able to clear it by rechecking group, then unchecking it again.  Others have had to reset the collar to clear it.

MNight75:
The person reporting it to me had just updated a brightcollar to the newest OC, doesn't use group and found group restrictions set that were not cleared by checking and unchecking group. 
Additionally a second person I helped with the issue tried the same thing and also failed, until the removed and replaced the collar. Then and only then would clicking and unclicking group clear the restriction.

Medea-Destiny:
I have a fix for this which I will be appearing in a pull soon. Basically the problem was that the restriction was being set if group key was not NULL_KEY, but group key gets set to "" if it's NULL_KEY so it needed to be set if group key was not "" instead. 

It has been suggested to me that an additional aspect to this is that the group access option can be set when there is no active group. The fix as above resolves the setgroup issue, but retains the misleading presentation of group access being set when there is no group value stored. To resolve this, my edit will disallow setting group access if there is no current group set.  Any comments/disagreements about this?

Medea-Destiny:
Fix provided in https://github.com/OpenCollarTeam/OpenCollar/tree/Medea-4.1-1

SilkieSabra:
I think that makes perfect sense (late response),

Trinkitz:
This seems to work as intended. Typo in the notice I receive in local chat when trying to set group access with no group active: " Ping Designs: roup access can't be set while no group is active."

Medea-Destiny:
Leaving this open to deal with typo issues for 8.2

Issue #562, opened 2021-05-07
Access not allowed for adding or removing owners, trusted, or blocking even with my owner wanting to give me permission to do so.


EliseBun:
What version of OpenCollar are you using?
8..5000
What behavior did you expect?
To have it so I, the wearer, am able to add trusted, and block others without any problems
What behavior did you see instead?
Each time I try to remove or add someone to owner, trusted, and block, it will give me" **RE** Simone Heart Collar v1.0: Access Denied to adding a person" or "**RE** Simone Heart Collar v1.0: Access Denied to adding a person" and I am not sure how to toggle it or give me access even with my owner wanting to have that available for me.
What steps does someone need to take to reproduce the problem?
First I download the new 8.0 patch on the collar and immediately my owner did the weld function. After so, I trying to add someone in trusted and I have no permission is to do so. I asked my owner to fix it so I can have access but even she could have not figure out how to do it, and also that her collar is also experiencing the same thing.

SilkieSabra:
The way to fix this is to add the wearer to the owner list.  It is expected behavior that only owners can change the access settings.  I do see room for a feature that allows owners to set a preference so that wearers can add trusted or set group without being able to change the owner access.

Medea-Destiny:
Fix provided in https://github.com/OpenCollarTeam/OpenCollar/tree/Medea-4.1-1

PR #592, opened 2021-07-04
Various fixes to oc_api and oc_core

Medea-Destiny:
oc_api:                 *Fix issue #566  setgroup not clearing properly, 
                            *Fix issue #562, #381, #495 allow owners to permit wearers to set trusted/block
                            *Fix issue #579 Add interface channel to DoListeners() function so Highlander works again
                            *Issue #579 Restored menuto function to interface channel for backwards compatibility
                            *Issue #579 Added control of channel 0 listening via settings menu
                            *Fix issue 585 give wearer accept/deny dialog to confirm runaway disable setting
oc_core:               *Fix issue #562, #381, #495 allow owners to permit wearers to set trusted/block.
                            -   added Wearer Trust option to Access Menu so that owners may allow or forbid
                                wearer from adding/removing to trusted and block lists
                            -   Moved Limit Range to Settings menu to make room in Access menu for above,
                                as it seems logical to have this as a global setting. 
                            *Added explanatory text to settings menu prompt and access prompt,
                            *Added safeword to help/about prompt. 
                            *Fix issue #566, clear @setgroup when group unchecked properly.
                            *Extention to above, Disallow setting group access when no group active.  
                            *Fix issue #580 Limited printing settings to owner and wearer
                            *#579 Added 'Listen 0' button to Settings menu that allows toggling channel 0 command listener on and off.  

These are issues as a single pull due to the interdependency of various functions of these scripts. Testing should be done with BOTH oc_api and oc_core at the same time.


Pingout (commented):
When I swap this script into the 8.0.5 collar I can not add an owner or trusted to my access. I get a sub menu when I try but the only option in the sub menu is to add >Wearer<
When both new oc_core and oc_api are in the collar, I get nothing when I try to add an owner or trusted. As Trinkitz says, the menu poofs

Trinkitz:
I cannot add an Owner, Trusted or Block with this new script. I tested everything I could, but until access is fixed, I cannot go any further with testing. Thank you.

Medea-Destiny:
That should fix the issue, I had a comparison check happening inside another comparison check that contradicted it. I will never be persuaded that K&R indentation style is a good idea - white space is free, readability is not. :P

Trinkitz:
We made a lot of progress today; forwarding findings to Ping. We still have some work to do, but so far this looks very good.

SilkieSabra (approved):
All of this looks good to me, thanks to our testing team! One wrinkle, I don't know git that well and this is asking to merge into master, it needs to go into https://github.com/OpenCollarTeam/OpenCollar/tree/features-2021-q2 instead.  I'm guessing it works if you are in the features branch when making the fix? I can't see how to merge into the correct branch on this pull.

Medea-Destiny:
USE #593 rather than this pull request!

I have no idea why this now wants to merge tayaphidoux's work on the AO stuff, this is possibly because I shifted to the features base from the master base and there's stuff that got merged with master rather than features and something got out of sync, but this ain't my stuff and I have no idea what the correct versions are so best to avoid.

PR #601, opened 2021-07-09
Update oc_remote.lsl

tayaphidoux:
Should Fix the Remote HUD disconnecting because of known reasons.


tayaphidoux:
committed to features instead of master as requested.

Issue #528, opened 2021-03-31
Undress not showing many system layers - Tattoo or Universal or Alpha


melanuss:
What version of OpenCollar are you using? 8.0.5

What behavior did you expect? I expected the blue menu under Apps>Undress>Rm Clothes to show ALL the system layers I was wearing: 
Jacket, Pants, Shoes, Socks, 
Alpha mask, Tattoo, Universal, Physics

What behavior did you see instead? The blue menu only showed Jacket, Pants, Shoes, Socks 

What steps does someone need to take to reproduce the problem?
In SL I wore fresh OpenCollar from OpenCollar Headquarters to test on myself. I was wearing system layers & parts (Skin, Tattoo, Universal Layer, Alpha, Pants, Jacket, Socks, Shoe,  eyes, hair base, Physics) plus items worn on attach points (hair, shoes, shirt, BOM mesh body). There are other system layers I was not wearing and did not test. I am reporting the things I was actually wearing.

zontreck:
Detach should be used for attachments.  Undress only shows the SYSTEM legacy clothing layers that would appear on baked on mesh. Attachments are separate

Mar 30, 2021 11:33:16 PM melanuss ***@***.***>:

melanuss:
Yes, attachments are separate from layers. 

I was testing both Detach and Undress and included the information about the attached items for completeness. I was wearing a mix of things on both system layers and attach points. Detach seems to be working fine and the way it lists items is a big improvement over the old RLV legacy Un/Dress I remember from OpenCollar v3 of some years ago. 

Undress should show ALL the system clothing layers. And that includes things that are not actually BOM baked textures. For example Physics.

melanuss:
I was discussing this with someone and they brought up the point that they are not sure if this is sort of in between a bug report and a feature request :-) 

Either way I took a moment to look up the entire list of New Clothing System Layers types that are in the LL inventory drop down menu. Hopefully this is helpful. They are the following 13 types: Shirt, Pants, Shoes, Socks, Jacket, Skirt, Gloves, Undershirt, Underpants, Alpha Mask, Tattoo, Universal, Physics. 

If it is something that can be detached or unworn I would owners to be able to manually remove it independently of whether of not outfits are properly set up in #RLV or not.

zontreck:
The undress app follows the RLV specification.  The viewer does not tell the script if you are wearing physics.  The script only displays what it is told by the viewer

[...]

PR #602, opened 2021-07-10
Update oc_undress.lsl

SilkieSabra:
restoring alpha, tattoo and universal layers, re issue #528


Medea-Destiny:
This is not the best way to deal with this issue due to differences between RLV and RLVa and potential future expansion of the wearables list in either.

Medea-Destiny:
See #604 and comments in #528

PR #603, opened 2021-07-10
Update oc_api.lsl

Medea-Destiny:
reinserting activation of interface channel lost in a prior revision


Pingout:
Thanks for the quick fix Medea!

PR #604, opened 2021-07-10
Update oc_undress.lsl

Medea-Destiny:
Extended layer handling and chat commands added

### DIFF
--- a/src/Apps/oc_undress.lsl
+++ b/src/Apps/oc_undress.lsl
@@ -351,8 +351,6 @@ state active
     
     listen(integer c,string n,key i,string m){
         if(c == g_iOutfitScan){
-            //llWhisper(0, "outfit worn reply: "+m);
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
This addresses #528 and adds chat commands to undress. I have done initial testing on Firestorm and it all seems good. However due to the inconsistencies in behaviour between RLV and RLVa this really should be testing on a different viewer that uses Marine's RLV implementation before merging

SilkieSabra (commented):
This makes sense to me.  Let's test.

Pingout:
Everything tested good. Works as expected.

Trinkitz:
We tested this with tattoo and clothing layers on a Maitreya mesh body. No problems; clothing lock also worked fine. We also removed the Maitreya body alpha using undress.

PR #608, opened 2021-07-12
Update oc_core.lsl

Medea-Destiny:
Limited changing public and group access setting to users with owner permission rather than owner or trusted


SilkieSabra:
This is restoring previous permissions which did not allow Trusted to alter Public or Group access settings.

Issue #525, opened 2021-03-29
Cuffs chain particle appearance 


Pingout:
What version of OpenCollar are you using?
OpenCollar 8.0.5 - OC Cuffs 8.0 Beta
What behavior did you expect?
The chains between cuffs should be more consistent as far as chain link appearance 
What behavior did you see instead?
Longer chains appear spread out and short chains appear to bunch up. A variable may be needed so long chains issue more particles than short chains. 

What steps does someone need to take to reproduce the problem?

Pingout:
I see this as a “nice thing to have” as opposed to a “needs immediate fix” issue

PR #614, opened 2021-07-17
Changes to Cuff Particle System, ref Issue #525

Pingout:
Safra has made some improvements to the chain appearance, and also has connected the cuffs to the collar's Leash Configure menu to allow color and texture change


Pingout:
Superseded

PR #616, opened 2021-07-19
Add Priority 6 Versions of the poses i made

lillith-xue:
I was told this may help with the cuffs.
Tell me if i missed some.


SilkieSabra:
This is great that Lill wants to contribute, Safra Nitely asked her to make a set of priority 5 animations which cannot be overridden.  I have qualms because of things like the posture anim and ankle fixes.  However RR animations appear to be priority 6 also.

SilkieSabra (commented):
Thank you Lill.  We will pull these as soon as we've checked to make sure they are not duplicates.  Also thanks for creating a contributor agreement page for yourself ! Very important as an example down the road.

SilkieSabra (approved):
I'm so confused ! Do we have these already at priority 6 or don't we?

PR #618, opened 2021-07-21
Changes to oc_cuff Re: particle system

Pingout:
Safra has added a method of changing the look and feel of the cuff chains using the Collar's Leash Configuration. Cuff chains will assume the Leash particle color and appearance

### DIFF
--- a/src/cuffs/oc_cuff.lsl
+++ b/src/cuffs/oc_cuff.lsl
@@ -5,14 +5,13 @@ Copyright ©2021
 : Contributors :
 Aria (Tashia Redrose)
     * February 2021       -       Created oc_cuff
-Safra (Safra Nitely)
-    * June 2021           -       add priority for animations, fix visual lock/unlock
 et al.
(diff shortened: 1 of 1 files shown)


Pingout:
There’s an issue with the particles. The chain particles look good on short chains, but medium length and longer chains appear stretched out. It’s hard to recognize the particles as chain texture. We will continue to test and maybe try some mods

tayaphidoux:
i hope line 8 and 9 are because safra's name was not in the source you uploaded and not because you are trying to remove it on purpose, since safra don't have a git hub her contributions should be documented this way at the very least.

Pingout:
> i hope line 8 and 9 are because safra's name was not in the source you uploaded and not because you are trying to remove it on purpose, since safra don't have a git hub her contributions should be documented this way at the very least.

Thanks for pointing out that Safra left her name out of this version. What's been happening is that Safra hands me a script and I usually have to put her name in there. We're trying to combine Safra's and Liliths changes into one update, so this branch will be deleted in favor of a more complete update

Issue #624, opened 2021-07-26
Cuffs pre-release notice remains in menu text of full cuffs release v1


Phuk-SLAM:
OpenCollar Cuffs 1.0.0

heyo OC team!
just a quick pointer, that the menu text of the new cuffs scripts still shows
** You are using a pre-release version. Some bugs may be encountered! **
while eventual bugs are something we all gotta embrace, this will potentially
confuse end-users. being a design partner, i'm hesitant to release the cuffs
as long as that notice is there. thx for looking into it, if possible!

Pingout:
This is my mistake I should have checked this. We should not be getting this error based on the following (which is not working correctly):

integer g_iAmNewer;

Compare(string V1, string V2){
    NEW_VERSION=V2;

    if(V1==V2){
        UPDATE_AVAILABLE=FALSE;
        return;
    }
    V1 = llDumpList2String(llParseString2List(V1, ["."],[]),"");
    V2 = llDumpList2String(llParseString2List(V2, ["."],[]), "");
    integer iV1 = (integer)V1;
    integer iV2 = (integer)V2;

    if(iV1 < iV2){
        UPDATE_AVAILABLE=TRUE;
        g_iAmNewer=FALSE;
    } else if(iV1 == iV2) return;
    else if(iV1 > iV2){
        UPDATE_AVAILABLE=FALSE;
        g_iAmNewer=TRUE;

        llSetText("", <1,0,0>,1);
    }
}

SilkieSabra:
Resolved ! thanks everyonee

PR #625, opened 2021-07-26
Update to remove pre-release notice Ref #624

Pingout:


SilkieSabra (approved):
minor update, been tested

PR #629, opened 2021-07-26
fix random resets on teleport,login.

tayaphidoux:
fix the script resets when teleporting or log in so that addon specific configurations can be saved in script.


Pingout (approved):
Thank you Taya!

Pingout (approved):

SilkieSabra:
Ping says it's ready to merge so here goes

PR #635, opened 2021-08-02
Add basic Lockmeister and add Lockguard Ping

lillith-xue:
Add basic Lockmeister implementation
Add Lockguard ping response for devices who check the available protocols before sending commands


lillith-xue:
It uses the list that was parsed out the notecard (g_lLMV2Map). So no additional points.
in line 990 it tries to find the lockmaster name and in line 991 it takes the mapped prim name and uses the GetKey function to get its key. In line 992 it checks if the name belongs to this cuff, and only if so it responds.

Basicly this only adds lockmeister responses. The cuffs where not able to respond to lockmeister in any way before. Tested that with my own devices and the npose chain demo device.

Pingout:
> 
> Basicly this only adds lockmeister responses. The cuffs where not able to respond to lockmeister in any way before. Tested that with my own devices and the npose chain demo device.

Very nice, thank you! We'll put it through some checks. I was just assuming the lockmeister was working, because lockguard worked so well and we never found any furniture it didn't work on. Thanks for fixing the missing piece!

lillith-xue:
Yes lockmeister is old, but there is still furniture out there that can not do anything else :-) For all others i also added the ping response to lockguard, so they see there is a lockguard implementation in the cuffs if they want to give lockguard priority

Pingout (approved):
This works great, thank you Lilith and Safra, also tested by Omnia, Trinkitz and Crysol

Pingout:
We narrowed down the disappearing leash to this script. What's happening is when a person wearing the cuffs trys to grab another person's leash, the leash quits after about 3 or 4 seconds.

PR #640, opened 2021-08-06
Safra and Lilith updates combined

Pingout:
Safra was kind enough to merge Lilith's Lockmeister functionality into her current working (Probably version 2.0) script.  This is an ongoing effort by Safra to make some needed fixes to the cuff script, including:
1) Fixed particle engine (v2 - with two weeks testing on over 30 devices)
2) New Chain and Ribbon textures - made by Safra. Tested by Safra with 2 weeks use on over 30 devices.
3) New texture engine call system to detect and activate the new cuff configuration system
4) Fixes to owner level detection and reduced calls on the config channels (reduced spam by 40%)
Note Particle rate reduction by 80% while still looking better than old.

### DIFF
--- a/src/cuffs/oc_cuff.lsl
+++ b/src/cuffs/oc_cuff.lsl
@@ -4,10 +4,16 @@ This file is a part of OpenCollar.
 Copyright ©2021
 : Contributors :
 Aria (Tashia Redrose)
-    * February 2021       -       Created oc_cuff
+    * February 2021         -       Created oc_cuff
 Safra (Safra Nitely)
-    * June 2021           -       add priority for animations, fix visual lock/unlock
+    * June 2021             -       add priority for animations, fix visual lock/unlock
+    * July 2021             -       added new prarticel engine - new chain and ribbon particles
+    * August 2021           -      addedd new texture control system. external script (oc_cuffs_config)
+ lillith (lillith Xeu)
+    * August 2021       -      fixed lockmeister system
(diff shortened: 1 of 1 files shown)


Pingout:
Re Lilith fix #635

Pingout:
We have two major changes to this Version 2 (Alpha) script. 
The first change connected the chain visual appearance to the Collar's leash configuration. So to change the appearance of the cuff chains, you go into the collar leash settings. The problem we have seen with that is we have the default leash is a ribbon texture, so everyone will want to know how to change it to a chain. Also, some people may want a different texture for cuffs than they can get for the leash
The second change was precipitated by the first. The leash configuration settings required a rework of the cuff particle generator, slowing it down, and changing the default texture uuid's with new chain textures.  This creates a few problems with the cuffs: First, the particles generate so slowly that if you are walking while wearing handcuffs, you leave a stream of particles behind you for 5 meters. Secondly the particles can not be seen if you are more than 10 meters from the person wearing the cuffs.  Also in the poses requiring long chains, for example the "Display" pose, the arm to leg chain is really stretched out. This stretched out chain can also be seen on any furniture that calls for chains longer than a meter. And the chain textures seem to have no thickness, one face looks correct, but then turn your view 90 degrees and the chain virtually disappears.

SilkieSabra (commented):
We need to know why the changes were made and what they were meant to accomplish.

Issue #642, opened 2021-08-11
Updating Cuff Arms and Legs configuration note cards


Pingout:
OpenCollar Addon Cuffs: Version 1.0

Because Lilith has been good enough to post her updated (Priority 6) cuff poses, we need to update Arms and Legs notecards to reflect the new poses.

I propose we rename each of Lilith's Priority 6 poses and use the P6 designation to differentiate the new poses from the older. I will attach updated Arms and Legs Notecards with the new pose names.

New priority 6 poses along with updated notecards will be ready for the next update.

PR #621, opened 2021-07-22
Cuff-Poses redone

lillith-xue:
Some Cuff-Poses where there is no source on GitHub. They are redone from scratch. Hopefully looking a bit better.

PR #643, opened 2021-08-11
Update Arms.txt

Pingout:
Reference Issue #642 and Pull Requests #616 and #621


Pingout:
This is under review, again trying to change pose names to save memory

PR #644, opened 2021-08-11
Create Legs.txt

Pingout:
Reference Issue #642 and Pull Request #616 and #621


Pingout:
This is under review. We are looking at shortening the pose names to save memory. Stay tuned!

PR #651, opened 2021-08-30
Remove the keying of the Pelvis in cuff pose

lillith-xue:
Fix pose "Lock Arms Front Close Prio 6" animation, locking pelvis bones.


Pingout (approved):
This one works much better, thanks for the fix!

Issue #652, opened 2021-09-04
oc_dialog fails to actually chat out long prompt text when promised.


Medea-Destiny:
When you send a menu to oc_dialog that has a prompt text that exceeds the 512 byte limit for llDialog prompts, you'll see the following message in the menu prompt: "Please check nearby chat for button descriptions."

This is a LIE. The script does parse the prompt text out to a variable to chat, but then forgets to actually chat it in nearby - we seem to have lost that snippet somewhere along the line.

While I'm poking I notice a string truncate function to truncate string length by byte length which can make for some very slow menus when prompt length exceeds the llDialog limit, and doesn't actually produce a very accurate result. I've got a replacement function for this which produces no appreciable delay to do a 1024 byte truncate compared to 24 seconds for the current function in my testing.

Medea-Destiny:
I have a fixed version of the script. Rather than pasting the old chat function back in, I've done a slightly nicer one that handles strings of any length, and will not break mid-word.

The new truncate function saves 510 bytes in the script bytecode when not running. In testing at a 1024 byte trim, the old function took 27 seconds while the new function took 0.12 seconds, and peak script memory usage was about 6kb lower.

PR #655, opened 2021-09-08
Update oc_dialog.lsl

Medea-Destiny:
Fixes for #652 and other improvements.

Changes: - Optimised substitutions by shortcutting no match for % in string after %NOACCESS% and %WEARER%
              - Replaced GetTruncatedString function with far more optimised function
              - Restored dialog chatting over-long prompts in local, which seems to have got lost somewhere. Improved
                version of function now works with prompts of any length and does not break up words.
              - Changed prompt for button descriptions in local chat from "Please check..." to "See..." to trigger it less often
              - Added warning text to prompt when prompt is trunctated, so menu prompt text will now show "(CONT. IN LOCAL CHAT)"

### DIFF
--- a/src/collar/oc_dialog.lsl
+++ b/src/collar/oc_dialog.lsl
@@ -3,7 +3,16 @@
 // Cleo Collins, Satomi Ahn, Joy Stipe, Wendy Starfall, littlemousy,
 // Romka Swallowtail, Garvin Twine, and Tashia Redrose,  et al.
 // Licensed under the GPLv2.  See LICENSE for full details.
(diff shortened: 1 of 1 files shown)


nirea:
I see this PR is targeting master.  Can you target it against the Q3 feature branch instead so the testers can test all the changes together before we bless the branch and release it?

Medea-Destiny:
> I see this PR is targeting master. Can you target it against the Q3 feature branch instead so the testers can test all the changes together before we bless the branch and release it?

Done

PR #658, opened 2021-09-08
RLV_CMD_OVERRIDE

Medea-Destiny:
This change adds a new command call to oc_rlvsys utilising the link message constant RLV_CMD_OVERRIDE to request a one-shot (=force) RLV command be applied overriding current restrictions. It is intended to be used only in the case where the person issuing the command has OWNER level auth.

The logic of this is simple: restrictions are intended to restrict the wearer of the collar, not an owner. If an owner wants to issue a command that conflicts with restrictions, they have the option of removing the restriction, issuing the command, and re-asserting the restriction. This simply automates the process. 

Usage: The function can be called by other scripts in the format `llMessageLinked(LINK_SET,RLC_CMD,OVERRIDE,commands+"~"overrides,kID); `
where 'commands' are a comma separated list of RLV commands that are to be issued with the "=force" (one shot) parameter, and 'overrides' are a comma-separated list of restrictions which will be compared to the list of baked restrictions, and if any of those restriction are in place they will be lifted before the one shot commands are issued, then reasserted afterwards.
For the sake of clarity, the person issuing the command (kID) will be notified of which restrictions were temporarily lifted when performing a command. This ensures that they are not misled into thinking a restriction is not in place.

EXAMPLE: Here's how this function can be used in the case of the UNSIT button being pressed:

```
 if(iNum==CMD_OWNER) llMessageLinked(LINK_SET,RLV_CMD_OVERRIDE,"unsit~unsit",kID);
 else llMessageLinked(LINK_SET,RLV_CMD,"unsit=force","Macros");
```

If the command issuer's auth level is CMD_OWNER, the RLV_CMD_OVERRIDE command is sent, requesting the unsit be temporarily lifted, and then an unsit=force command be issued. If auth is not CMD_OWNER, the current behaviour - a standard RLV_CMD message is sent. In the former case, if an unsit restriction is in place, the command issuer will receive the message "Owner Override. Restiction(s): unsit -- were temporarily lifted to perform request, but are now back in place. In the latter case, the command issuer would receive the message "Can't force stand due to RLV restrictions".

Without this feature being used, Owners would receive the "Can't force stand due to RLV restrictions" message. They would then have to navigate to the restrictions menu, release the unsit restriction, navigate back to the force sit menu, issue the unsit, then navigate back to the restrictions menu again, to re-state the restriction.

CONSEQUENCES of this change: 
The extra lines of code in RLVsys. **This change has no impact on any currently implemented features**, it simply allows us to create a more user-friendly way of doing things in future.

### DIFF
--- a/src/collar/oc_rlvsys.lsl
+++ b/src/collar/oc_rlvsys.lsl
@@ -1,8 +1,17 @@
-// This file is part of OpenCollar.
-// Copyright (c) 2008 - 2016 Satomi Ahn, Nandana Singh, Wendy Starfall,
-// Medea Destiny, littlemousy, Romka Swallowtail, Garvin Twine,
-// Sumi Perl et al.
-// Licensed under the GPLv2.  See LICENSE for full details.
+/*
+ This file is part of OpenCollar.
+Copyright (c) 2008 - 2016 Satomi Ahn, Nandana Singh, Wendy Starfall,
+Medea Destiny, littlemousy, Romka Swallowtail, Garvin Twine,
+Sumi Perl et al.
+Licensed under the GPLv2.  See LICENSE for full details.
+
+Medea Destiny   -
+        Sept 20201  -   Added RLV_CMD_OVERRIDE function. This allows one shot (=force) commands to be sent that will override any restrictions. 
+                        This command should only be used where operator has owner permission. The notion behind it is that owners should
+                        not be restricted by wearer restictions. They can unset restrictions manually, perform the function, and then reset
+                        the restriction, but that's a lot of hassle. This performs the function automatically. Operator and wearer are notified of 
+                        restrictions that have been temporarily restricted to avoid being misled that a restriciton is not present.
+*/
 
 string g_sScriptVersion = "8.1";
 integer g_iRLVOn = TRUE;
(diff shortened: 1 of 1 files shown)


nirea:
I see this PR is targeting master. Can you target it against the Q3 feature branch instead so the testers can test all the changes together before we bless the branch and release it?

src/collar/oc_rlvsys.lsl
```diff
@@ -620,7 +630,31 @@ state active
...
+                list lOverrides=llParseString2List(llList2String(llParseString2List(sStr,["~"],[]),1),[","],[]);
+                integer iLen=llGetListLength(lOverrides);
+                while(iLen--) {
+                    if(llListFindList(g_lBaked,llList2List(lOverrides,iLen,iLen))==-1)
+                    {
+                        lOverrides=llDeleteSubList(lOverrides,iLen,iLen);
+                    }
+                    else llOwnerSay("@"+llList2String(lOverrides,iLen)+"=y");
+                }
+                llSleep(0.5);
```
nirea, line 645: What are we waiting for here?

Medea-Destiny: This is between temporarily releasing restrictions that will block a one-shot RLV command and issuing that one-shot command. The sleep is to give the release a chance to take effect before attempting the one-shot command. Without it, we could easily be sending the viewer a command to release a restriction and the command to issue the one-shot in the same frame and I wouldn't want to assume that the viewer will process these in the correct order in that case.

PR #659, opened 2021-09-08
Update oc_unwelder.lsl

SilkieSabra:
Improved the instructions to uncheck AddOns Limited before proceeding with the unweld.

### DIFF
--- a/src/spares/oc_unwelder.lsl
+++ b/src/spares/oc_unwelder.lsl
@@ -218,7 +218,7 @@ default
                     
                     if(sStr == "settings=sent"){
                         if(g_iAddonLimitation){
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
@SilkieSabra This looks like a nice, user-friendly change! One thing though:

"To unweld, a collar owner must go to the collar Settings menu and find the AddOn Settings button"

I've checked the relevant menu in oc_core and permission to make this change is for OWNER or TRUSTED. I'd imagine in a good proportion of cases but not all, the user will have recently run away, and might not be aware that by default they will now have OWNER permission. Can we have a slight rewording that would clarify this?

Beyond the scope of this PR,  but I notice the welder uses llSay to chat this message (and a number of others) in local chat rather than directing it at the user. Is this something we should consider changing to avoid spamming local?

Medea-Destiny (approved):
Changes made to add clarity look good to me. Let's merge!

Issue #630, opened 2021-07-27
Cuffs Restrictions don't release


SilkieSabra:
In OpenCollar Addon Cuffs V1, we set restrictions for each pose with the line PoseRestrictions:[restriction]=[value].  The restrictions were not individually tested for V 1.0.  It now turns out that the line will set the restriction, but the restriction does not clear when the pose is changed or stopped, until after a relog.

So, if each pose has a set of restrictions, and you change poses, the new restrictions are added and the restrictions from the former pose are retained.  If you stop the pose, the restrictions are still enforced.  Relogging clears it.

Pingout:
See PR #662 for the possible fix

PR #662, opened 2021-09-15
Update oc_cuff.lsl

Pingout:
This is a major update which includes fixes by Safra and Lilith.
Particle system has been updated, Pose restriction error fixed, and LockMeister functionality added.
Testing team has this script in the White/Red set


Pingout:
The pose restriction problem Issue #630 was fixed by just removing the comment marks on line number 909 and 918, here's the logic:
Before Safra's access levels, each time a pose changed, regardless of who changed it, the script would clear restrictions

So their fix at the time was to make ClearChains an owner only option, and comment the lines that had the script do it automatically

which locked out everyone else from removing restrictions

Now with Safra's auth system in place, the only way to reach that "Link("from_addon", CLEAR_ALL_CHAINS, "", ""); " command in the script is if you first had enough auth level to set the pose & restriction

With out that auth level you get a pose button that does nothing

ClearChains should remain owner only, as it is a god-mode power that does a Stop and Clear for poses particles and restrictions, yet retains the detach=n necessary for locked cuffs.

nirea:
I see this PR is targeting master. Can you target it against the Q3 feature branch instead so the testers can test all the changes together before we bless the branch and release it?

Pingout:
> I see this PR is targeting master. Can you target it against the Q3 feature branch instead so the testers can test all the changes together before we bless the branch and release it?

In general, the Cuff Scripts have been on a separate trajectory from the Collar Script Quarterly Updates. We have been merging them directly to master once testing is completed

Pingout:
The testing team has turned up what appears to be a bug. The bug happens when the person wearing the cuffs attempts to grab the leash of another collar wearer. The Leashed collar spawns particles for maybe three or four seconds then the emitter stops working correctly. 
* It seems to make no difference what collar version the Leashed avatar is wearing. 
* The leash effect is still in place, for example the Leashed avatar follows the Leasher. 
* On teleport, the effect happens again. About 3 or 4 seconds after the Leashed avatar is teleported, the particle emitter stops. 
* I have noticed every minute or so, the collar emitter will spit out a handful of particles, 4 or 5 at most, and stops emitting again.
* We were not able to reproduce this bug on any other Collar/Cuff combination, it seems to be limited to this version of the cuff script

Since this script is a major upgrade of the Cuff particle system which uses the collar's Leash Configuration to change Cuff chain appearance, there may be some sort of conflict created which only appears when Cuffs and Leash are both in use.

I am marking this PR as "Do Not Merge"

Thanks Trinkitz for uncovering and reporting this bug

Medea-Destiny:
@lillith-xue @safra 

Re lines 1018 to 1026: 
```
 if (sLMCMD == "RequestPoint")
 {
        key kLink = NULL_KEY;
        list lKey = [];
        integer iMapIndex = llListFindList(g_lLMV2Map, [sLMPoint]);
        if (iMapIndex > -1) lKey = GetKey(llList2String(g_lLMV2Map, iMapIndex + 1));
        if (llList2Integer(lKey, 0) != LINK_ROOT)
        llRegionSayTo(id, -8888,(string)llGetOwner()+"|LMV2|ReplyPoint|"+sLMPoint+"|"+llList2String(lKey, 1));
}
```

This seems to be the cause of the issue @Pingout referenced above. The collar sends an LMV2 requestpoint for collar as part of the leash holder handling, and the code here does not handle a case where the point requested is not in g_iLMV2Map. Note that if iMapIndex is -1 it will still execute the code on the line below -- the empty list lKey will not have a first element equal to LINK_ROOT, so for unfound leash points it's sending a replypoint with an empty string for the key, which the collar's oc_particle script will set as the particle target, which means the collar's leash particles stop. I suggest changing the code to:
```
 if (sLMCMD == "RequestPoint")
 {
        key kLink = NULL_KEY;
        list lKey = [];
        integer iMapIndex = llListFindList(g_lLMV2Map, [sLMPoint]);
        if (iMapIndex > -1) 
        {
               lKey = GetKey(llList2String(g_lLMV2Map, iMapIndex + 1));
               if (llList2Integer(lKey, 0) != LINK_ROOT)
                     llRegionSayTo(id, -8888,(string)llGetOwner()+"|LMV2|ReplyPoint|"+sLMPoint+"|"+llList2String(lKey, 1));
         }
}
```

safra-nitely:
I have added this fix into the current in-world build of the cuffs.
I have also added a builders kit into the box, to help people build there own set.  Needs a notecard to explain naming convetions of the leashing points.  I may do this alter today. 
2.2.1.e - and sent it out copied of the update to both ping and roan.

Omnia2021 (approved):

Omnia2021 (approved):

Omnia2021 (approved):

Pingout:
Testing has turned up a problem in the Cuff auth priority system. Here's how it should work:
Someone who is Wearer, Trusted or Owner can set a pose on the cuff wearer. The Wearer and Trusted person should be able to stop or change the pose they set, but the Wearer can not stop a pose set by Trusted, and Trusted can not stop or change a pose set by an Owner. Owner is all powerful, able to stop or change any pose regardless of who set the pose. This is the auth priority system Safra added to the cuffs.
The problem happens when the pose is set by either the Wearer or Trusted. The problem is the pose menu does not display when trying to enter the menu a second time. As long as the Wearer or Trusted keeps the pose menu open, they are able to changes or stop poses. But if they close the pose menu and later try to re-open that pose menu, the get no menu. There is also no message to the person trying to open the pose menu.
Of course, Owner being all powerful has no such issue trying to re-open a closed pose menu. 
I have placed a Do Not Merge tag on this PR until we can address this issue.

tayaphidoux:
didn't some one else make authorization changes to the cuffs related to addon authorization changes?

Pingout:
> Testing has turned up a problem in the Cuff auth priority system. Here's how it should work: Someone who is Wearer, Trusted or Owner can set a pose on the cuff wearer. The Wearer and Trusted person should be able to stop or change the pose they set, but the Wearer can not stop a pose set by Trusted, and Trusted can not stop or change a pose set by an Owner. Owner is all powerful, able to stop or change any pose regardless of who set the pose. <snip> I have placed a Do Not Merge tag on this PR until we can address this issue.

This appears to be a typo from when I copied the script from github to inworld. I think what I would like to do, since this PR is getting long, and convoluted, and frankly confusing, I would like to close this PR and open a new one with our current inworld working version.

Omnia2021:
Cuffs seem to work well, no major issues, start up is quick and the cuffs respond as intended

Issue #665, opened 2021-09-19
add-ons menu auth fail produces no response.


Medea-Destiny:
Per 8.1, if someone on public access (CMD_EVERYONE) clicks the add-ons menu button, they fail auth check without any response from the collar.

The user should be notified that they do not have sufficient access and returned to the main menu rather than getting no feedback.

PR #666, opened 2021-09-19
Update oc_addons.lsl

Medea-Destiny:
Provides fix for issue #665 

Do not merge pending tests.


tayaphidoux:
thats basicaly what i have done to fix the lock function on my items.

Medea-Destiny:
Auth was being checked twice, blocking the auth failure notification actually going through. Updated to fix.

PR #667, opened 2021-09-19
Boot process speedup

Medea-Destiny:
These changes make the boot process / timings much more aggressive. 

I have been running with these changes for 2 weeks now without issue, and have passed them on to our testers to play with. My estimate is that it roughly halves the boot time. @Trinkitz feels this might be conservative. 

These changes will mean the boot process is less bullet-proof, so thorough testing over time and in particularly laggy sims, of login, attaching and fix menus function should be done before okaying this, but I have yet to experience a single problem.

### DIFF
--- a/src/collar/oc_settings.lsl
+++ b/src/collar/oc_settings.lsl
@@ -1,19 +1,19 @@
-
 /*
 This file is a part of OpenCollar.
 Copyright ©2021
-
-
 : Contributors :
 All contributors of previous revision of oc_settings
-
 et al.
 Licensed under the GPLv2. See LICENSE for full details.
 https://github.com/OpenCollarTeam/OpenCollar
 
+Medea Destiny
+        Sept 2021   -   Tightened timings on release of all settings and reduced sleep
+                        padding on initialize
 */
 
 //string g_sParentMenu = "Apps";
+
 
 integer TIMEOUT_REGISTER = 30498;
 integer TIMEOUT_FIRED = 30499;
@@ -509,7 +509,7 @@ default
                 g_iBootup=FALSE;
             }
             llSetTimerEvent(0);
-        }else llSetTimerEvent(0.25); // send 1 setting per quarter second
+        }else llSetTimerEvent(0.15); // 0.25 send 1 setting per quarter second
     }
 
 
@@ -598,7 +598,7 @@ default
                 llSleep (5); // Sleep for 5 seconds to give some padding for all scripts to switch to the ready state!
                 g_iBootup=TRUE;
                 g_iCurrentIndex=0;
-                llSetTimerEvent(5);
+                llSetTimerEvent(0.5); //5
             }
         }
         //llOwnerSay(llDumpList2String([iSender,iNum,sStr,kID],"^"));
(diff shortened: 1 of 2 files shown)


src/collar/oc_settings.lsl
```diff
@@ -509,7 +509,7 @@ default
                 g_iBootup=FALSE;
             }
             llSetTimerEvent(0);
-        }else llSetTimerEvent(0.25); // send 1 setting per quarter second
+        }else llSetTimerEvent(0.15); // 0.25 send 1 setting per quarter second
```
nirea, line 512: Let's either get this comment in sync with the code or make it more general so we don't have to keep it in sync.

src/collar/oc_states.lsl
```diff
@@ -187,35 +184,35 @@ default
...
-            } else if(llGetTime()>=7.5 && g_iPasses>=3){
+            } else if(llGetTime()>=2 && g_iPasses>=2){ //>=7.5
                 if(g_iVerbosityLevel>=2)
                     llOwnerSay("Scripts ready: "+(string)llGetListLength(g_lAlive));
                 llMessageLinked(LINK_SET,STARTUP,llDumpList2String(g_lAlive,","),"");
                 g_iExpectAlive=0;
                 g_lAlive=[];
                 g_iPasses=0;
-                llSleep(10);
+                llSleep(1); //10
```
nirea, line 215: I know you didn't write it, but I _hate_ code with arbitrary sleeps to let other things finish.  It puts us in this horrible position of having to choose between waiting forever or risking that one of those things isn't done.

What are we waiting for here?  Is there a way we can get notified of that thing or check for it instead of just sleeping?

src/collar/oc_settings.lsl
Medea-Destiny: Done

src/collar/oc_states.lsl
Medea-Destiny: Essentially this is looping through a process where the scripts report back an alive status and leaving a cushion to ensure that every script has had a chance to do its own startup things before settings start getting sent out.

There's no straightforward way to receive notifications of this. As of now we're basically setting things running and moving ahead when there haven't been any new scripts reporting that they have come alive for a little while. Some kind of padding here is necessary unless we implement a method for every script to report not just an alive state but a ready state. This method isn't particularly scientific, but this tuning of the timing has worked so far in testing. 

We could beneficially redo the entire startup process, but that's a pretty major project.

Medea-Destiny:
Current changes boot the collar too fast to work smoothly with the updater. In collars with a lot of settings in place, they can get dropped. Two solutions:

1. Slow down the boot process again.
2. Fix the installer process.

I strongly favour 2. The installer is... I'm going to be polite and say problematic. It could do with a complete re-write honestly. However I have made a couple of small edits to oc_update_shim which means that it no longer re-requests settings multiple times causing an unnecessary flood of linked messages just because it hasn't received every setting within 5 seconds, which is hardly a surprise when oc_settings sends them with a quarter second delay and doesn't try rebooting the collar and sending the settings saved from the original oc_settings script to the new oc_settings script while the collar is going through the boot process, and this has so far proven entirely reliable. I have sent a copy to Trinkitz who's got a super-pimped out collar to try this on.

PR #670, opened 2021-09-21
Update oc_core.lsl

Medea-Destiny:
Added notifications when public or group mode is toggled.


Medea-Destiny:
New changes: inlined 3 functions that were being called a single time each to save script memory and added a small llSleep when changing device name so that the notification that the change has been made is issued AFTER the change has been made, rather than reporting the previous name.

PR #680, opened 2021-10-03
Update oc_titler.lsl

Medea-Destiny:
Reworking a few things in oc_titler as a result of report from @SilkieSabra of someone having trouble setting custom colour of text. Custom colour textbox did not give info on the required format, and if the input was not in the correct vector format would read the output as <0,0,0> thus setting text to black. 

I have added prompt text, and created a string to vector parser that will come up with a valid colour from a much wider range of inputs. This also fixes issues with chat commands. Additionally I did a drive-by efficiency improvement in saving settings.

Script notes:
                    Refactored SAVE to only save changed settings. Added helpful menu text.
                    Added text2col() function that makes a sensible colour vector out of whatever
                    is thrown at it. Added Added validvector() function to confirm vectors are valid
                    before setting, used to make "[prefix] title color" command without following vector
                    provide colour menu rather than setting colour to black, and means that "[prefix]
                    title color red" will now pop up the color menu rather than setting the title text
                    to black. Text changes now set instantly rather than after 2.5 second delay.

### DIFF
--- a/src/collar/oc_titler.lsl
+++ b/src/collar/oc_titler.lsl
@@ -160,12 +160,13 @@ UserCommand(integer iNum, string sStr, key kID) {
         if(iNum !=CMD_OWNER)return;
 
         if(sChangetype == "title"){
-            g_sTitle = llDumpList2String(llList2List(llParseString2List(sStr,[" "],[]), 1,-1)," ");
-            if(g_sTitle == ""){
+            string sTitle = llDumpList2String(llList2List(llParseString2List(sStr,[" "],[]), 1,-1)," ");
+            if(sTitle == ""){
                 Dialog(kID, "What should the title say?", [], [], 0, iNum, "Textbox~Title");
-
+                return;
             }
-            else Save(4); //only save if we're changing it
+            g_sTitle=sTitle;
+            Save(4); 
         } else if(sChangetype == "titler"){
             if(sChangevalue == "color"){
                 if(!validvector(sParam)) ColorMenu(kID,iNum);
--- a/src/collar/oc_titler.lsl
+++ b/src/collar/oc_titler.lsl
@@ -1,22 +1,28 @@
 /*
 This file is a part of OpenCollar.
 Copyright 2021
-
 : Contributors :
 Aria (Tashia Redrose)
     * Mar 2020      - Fix bug where title would not detect the float text prim.
                     - Fixed bug where titler would display garbled text as a result of failing to decode base64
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
Script has passed my testing but could use a bit more testing before merge.

Pingout:
This looks good Medea, thank you. I tried going back and forth from preset colors to custom colors and didn't have any problems. I also forced an error using numbers that were out of range and got this message: "Not a valid vector! Please try again." I got the same message when trying to submit an incomplete vector.

src/collar/oc_titler.lsl
```diff
@@ -166,27 +209,36 @@ UserCommand(integer iNum, string sStr, key kID) {
...
-
-    llMessageLinked(LINK_SET, LM_SETTING_SAVE, "titler_offset="+(string)g_iOffset, "");
-
-    if(!g_iNoB64)
-        llMessageLinked(LINK_SET, LM_SETTING_SAVE, "titler_title="+llStringToBase64(g_sTitle), "");
-    else
-        llMessageLinked(LINK_SET, LM_SETTING_SAVE, "titler_title="+g_sTitle, "");
-
-    llMessageLinked(LINK_SET, LM_SETTING_SAVE, "titler_color="+(string)g_vColor,"");
+Save(integer what){// 1= save showhide, 2-save offset 4=save title 8=save colour
```
nirea, line 212: Can you please define constants for these?

src/collar/oc_titler.lsl
```diff
@@ -263,7 +315,7 @@ state active
         NukeOtherText();
         //llOwnerSay((string)llGetUsedMemory());
     }
-    timer(){
+/*    timer(){
```
nirea, line 318: If we don't need a timer event anymore, let's remove it rather than leave this comment.

src/collar/oc_titler.lsl
```diff
@@ -342,26 +394,30 @@ state active
                         Menu(kAv, iAuth);
                         iRespring=FALSE;
                     } else if(sMsg == "Custom"){
-                        Dialog(kAv, "[Titler Custom Color]", [], [], 0, iAuth, "Textbox~Color");
+                        Dialog(kAv, "[Titler Custom Color]\nType in a custom colour using the R,G,B format, i.e. '255,0,0' gives pure red.", [], [], 0, iAuth, "Textbox~Color");
```
nirea, line 397: Love this.  Great clarification.

PR #685, opened 2021-10-09
Speedup bootup by responding to other script feedback.

Yosty7B3:
Overhaul of oc_states to go through distinct states and waiting for feedback from other scripts before moving to the next state. oc_settings is now sending out an ALIVE signal on REBOOT/script reset, when the .settings notecard is loaded.
This helps make the bootup process move on as soon as the settings are loaded rather then wait an arbitrary sleep time.
A failsafe timeout is included in case a oc_settings is used that does not send ALIVE.

default state:
*    send REBOOT reboot
*    waits for ALIVE from oc_settings, oc_core, oc_api

state startup:
*    send STARTUP
*    waits for :
  0 initialize from oc_core
  LM_SETTING_REQUEST ALL from oc_api
  "settings=sent" from oc_settings

state running:
*    send TIMEOUT_READY

The settings editor is moved to oc_settings because that's where the settings are stored and editing them directly makes it a lot more responsive.

oc_settings ALIVE & oc_states STARTUP also sends global_ settings for early access to settings like global_locked and global_verbosity

### DIFF
--- a/src/collar/oc_settings.lsl
+++ b/src/collar/oc_settings.lsl
@@ -15,6 +15,90 @@ https://github.com/OpenCollarTeam/OpenCollar
 
 //string g_sParentMenu = "Apps";
 
+
+integer g_iSettingsLoaded;
+integer g_iLastStride;
+string  g_sVariableView;
+string  g_sTokenView="";
+
+// return a CSV containing all global_var=val
+string SettingsFor(string SelectToken){
+    list lSettings;    
+    integer i = llGetListLength(g_lSettings); 
+    while((i-=2)>=0){        
+        list lTmp = llParseString2List(llList2String(g_lSettings,i),["_"],[]);
+        string sTok = llList2String(lTmp,0);
+        string sVar = llList2String(lTmp,1);
+        if(sTok == SelectToken)
+            lSettings+=llList2String(g_lSettings,i)+"="+llList2String(g_lSettings,i+1);
+    }
+    return llList2CSV(lSettings);
+}
+
+// sends "ALIVE oc_setttings <global_var=val list>"
+sendALIVE(){
+        g_iSettingsLoaded = TRUE;
+        llSetTimerEvent(0);   
+        llMessageLinked(LINK_SET, ALIVE, "oc_settings", SettingsFor("global")); 
+}
+
+// dialogs for "settings edit"
+SettingsMenu(integer stridePos, key kAv, integer iAuth){
+    string sText = "OpenCollar - Interactive Settings editor";
+    list lButtons = [];
+    if(iAuth != CMD_OWNER){
+        sText+="\n\nOnly owner may use this feature";
+        Dialog(kAv, sText, [], [UPMENU], 0, iAuth, "Menu~Main");
+        return;
+    }
+    integer end = llGetListLength(g_lSettings); integer i;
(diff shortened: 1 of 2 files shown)


src/collar/oc_settings.lsl
```diff
@@ -15,6 +15,90 @@ https://github.com/OpenCollarTeam/OpenCollar
...
+        string sTok = llList2String(lTmp,0);
+        string sVar = llList2String(lTmp,1);
+        if(sTok == SelectToken)
+            lSettings+=llList2String(g_lSettings,i)+"="+llList2String(g_lSettings,i+1);
+    }
+    return llList2CSV(lSettings);
+}
+
+// sends "ALIVE oc_setttings <global_var=val list>"
+sendALIVE(){
```
nirea, line 39: Can you capitalize this consistently with the other functions?

```suggestion
SendALIVE(){
```

src/collar/oc_settings.lsl
```diff
@@ -15,6 +15,90 @@ https://github.com/OpenCollarTeam/OpenCollar
 
 //string g_sParentMenu = "Apps";
 
+
```
nirea, line 18: I see this MR is targeted against master.  You might want to target it against the q3 feature branch to reduce the likelihood of merge conflicts later.

src/collar/oc_states.lsl
```diff
@@ -142,119 +80,167 @@ integer DIALOG_TIMEOUT = -9002;
...
+        g_lWaiting = llDeleteSubList(g_lWaiting, iScript, iScript);
+        //g_lAlive+=[scriptName];
+    }
+    if(llGetListLength(g_lWaiting)==0){ 
+        return TRUE;
+    }
+    return FALSE;
+}
+
+integer antiCrash(string scriptName){
```
nirea, line 107: Capitalization here too.

src/collar/oc_settings.lsl
nirea: It's hard to tell just from the code how these changes will alter the collar's behavior.  So my comments here are mostly about style.  @Medea-Destiny has been poking at startup times and hopefully has some more insight for you.

nirea (changes requested):

src/collar/oc_settings.lsl
Medea-Destiny: I have labelled this for targeting 8.3 as the changes here are significant and will take some solid poking at and testing. In particular moving settings edit to oc_settings while very logical is going to require some solid testing -- we try to keep oc_settings as small as possible because it's where we're storing an arbitrary amount of settings data. 

@Yosty7B3 This work is really appreciated and much needed stuff, many thanks for taking a look at it. It would be helpful to talk it over some time either in world or on Discord, let me know if you'd be okay with that.

src/collar/oc_states.lsl
Yosty7B3: done

src/collar/oc_settings.lsl
Yosty7B3: done

src/collar/oc_settings.lsl
Yosty7B3: How can I re-target this? Or should I do new pull request to target against q3?

PR #686, opened 2021-10-10
Q3 Style Tweaks

nirea:


src/collar/oc_addons.lsl
```diff
@@ -436,7 +436,7 @@ state active
         }
 
 
-        if(iNum >= CMD_OWNER && iNum <= CMD_EVERYONE) UserCommand(iNum, sStr, kID);
+        if(iNum >= CMD_OWNER && iNum <= CMD_EVERYONE) UserCommand(iNum, sStr, kID);// Used to be CMD_WEARER.  What's the effect of the change?
```
nirea, line 439: @Medea-Destiny I saw this changed in your branch from CMD_WEARER to CMD_EVERYONE.  I'd like to understand better what that does.

src/collar/oc_core.lsl
```diff
@@ -46,11 +46,6 @@ integer g_iVerbosityLevel=1;
 
 integer g_iNotifyInfo=FALSE;
 
-/*string MajorMinor(){
```
nirea, line 49: In order to keep the scripts clean over time for everyone, let's prefer deletion over commenting things out like this.

src/collar/oc_rlvextension.lsl
```diff
@@ -286,30 +282,32 @@ MenuSetValue(key kID, integer iAuth, string sValueName) {
 integer g_iLastOwnerEx;
 integer g_iLastTrustedEx;
 
-updateList(list newlist, integer type) // type 1=owner, 2=trusted;
+integer LIST_TYPE_OWNER = 1;
```
nirea, line 285: Naked magic numbers passed into functions can be a big impediment to readability.  We should use constants unless the script is actually crashing from running out of memory, and even then, this would be one of the last things to attempt to get some memory back.

src/collar/oc_rlvextension.lsl
```diff
@@ -286,30 +282,32 @@ MenuSetValue(key kID, integer iAuth, string sValueName) {
...
+integer LIST_TYPE_TRUSTED = 2;
+UpdateList(list newlist, integer type) ;
```
nirea, line 287: All the rest of the functions in this file (and all of OC) use `UpperCamelCase()`.  Please don't introduce a new style (like `lowerCamelCase` unless we've agreed to it as a team and written it down in a style guide that we can send new devs to.

src/collar/oc_addons.lsl
Medea-Destiny: In UserCommand that leads to the line:
` if(iNum>=CMD_OWNER && iNum<=CMD_WEARER) AddonsMenu(kID, iNum);`

So what happened is if you had a valid auth for the AddonsMenu it got checked twice, and if you didn't the collar was non-responsive to the click. 

In my edit we pass all valid users to the UserCommand and rely on filtering in the UserCommand to pass the AddonsMenu dialog to users with sufficient auth and present a no access error message to people who don't.

src/collar/oc_core.lsl
Medea-Destiny: My inclination is to leave a deleted function commented out for one revision for the sake of easily seeing what's been changed, but I'll keep this in mind in future. In this case as the commented functions have just been inlined into the body of the code to save memory it's not really an issue anyway.

src/collar/oc_addons.lsl
nirea: Thank you!

Medea-Destiny (approved):
Looks good to go

PR #689, opened 2021-10-15
Update oc_api.lsl

Medea-Destiny:
@Yosty7B3 noticed the the StrideOfList function in oc_api is unused. This edit simply removes the unused function to save a bit of memory.


SilkieSabra:
Can you merge this into q3 instead of Master, @Medea-Destiny

PR #694, opened 2021-10-31
8.2 version number

Medea-Destiny:
Update oc_core version number string to 8.2.1000 in readiness for release.


Pingout:
I'm always confused about version numbers. Shouldn't this be Version 8.2.0 ?

nirea:
I don't mind skipping ahead to 8.2.1.  I think the version we have been using for the inworld release candidate is 8.2.05 or something like that, and it's good for the released version to be higher to avoid confusion.

PR #695, opened 2021-11-01
2021 q3 features

nirea:


SilkieSabra (approved):
We've put this out for beta testing and got very few responses.  No news is good news and delaying release to try and get more participation in the beta is a loss leader.

Pingout (approved):
We have done quite a bit of testing and everything is working as expected. I like the layout of the RLV Restrictions menu. And customers have been begging for this faster start since we rolled out 8.0 beta.

PR #697, opened 2021-11-02
coment debug output for oc_rlvextension

Medea-Destiny:
Stray memory output debug message left uncommented.


SilkieSabra (approved):
Because no debug messages should be left uncommented

PR #699, opened 2021-11-02
Fix back button on muffle settings submenu

Medea-Destiny:
ARGH! Back button in muffle settings submenu doesn't work! This fixes.


Medea-Destiny:
Also back button on camera menu

Also updated the website link text in prompts to make sure it goes to the correct place.

Pingout (approved):
This works good inworld, the buttons are correct now. Thanks Medea!

Pingout (approved):
This works inworld. It's a lot of layers deep into the RLV Restrictions/Customize menu, but I found it and it works. :-)

PR #702, opened 2021-11-08
functions setor() and bool() removed

Yosty7B3:
To save script memory.

### DIFF
--- a/src/collar/oc_core.lsl
+++ b/src/collar/oc_core.lsl
@@ -166,18 +166,19 @@ AccessMenu(key kID, integer iAuth){
         return;
     }
     string sPrompt = "\nOpenCollar Access Controls\n+/- buttons to control access lists.\nGroup allows access to current group, Public allows access to all. Wearer trust allows an owned wearer to add/remove from trusted list.\nRunaway removes all owners, access list prints out who has access.";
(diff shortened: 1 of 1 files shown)


Medea-Destiny (commented):
Yosty7B3's changes here are intended to produce no user-space change, they are about doing the same thing in a different way. 

Replacing the bool function seems like a no-brainer to me, I see no advantage to the bool() method over checking the boolean inline. The setor() is a slightly more complicated question. In theory the setor() method should run in fewer cycles than an llList2String call, but at the cost of the overheads for declaring an extra function. Given it's not exactly being used in a lot of places I'm confident the replacement here makes sense. Perhaps most importantly, I think the setor() function is likely to be less transparent to new developers trying to make their way around this code. 

I'm not hitting merge on this yet because I want to be a bit more diligent by doing some efficiency testing, but this all looks good to me.

KristenMynx:
I like these changes.  Its more clear.   setor, a function with 3 parameters takes less slightly less memory to call, but it is offset by the extra function.   I see 2 references, anything gained is lost by obscurity.   I can't imagine a performance hit in any event, since these are menu and dialog initiated code sections.   Its not like they are running periodically nor even often.   I took the current oc_core, and applied the changes manually in-world, it resulted in 2000+ more free memory.    I will test targeting these changes and report.  I'll probably be suggesting a merge into 2022-q1-features.

Medea-Destiny:
Replacing this pull with #865 to make merging easier.

Issue #715, opened 2021-12-15
Teleport Function for Collar Owner


FunViking:
What version of OpenCollar are you using? 8.2.1000

What behavior did you expect? To be able to teleport to same location as my sub with a single click when I force teleport my sub. 

What behavior did you see instead? Unable to teleport to same location as my sub after I force teleport, without having to wait until sub arrives at location and sends me teleport request.  There was a function on the peanut 9 collar that popped up after I force teleported my sub which asked me if I wanted to teleport to same location.  A popup menu or HUD came up as I sent my sub and all I had to do was click yes or no.  We would arrive at the location at about the same time. This function was very beneficial when I would take her shopping or exploring.  

What steps does someone need to take to reproduce the problem? There is no problem I am just making suggestion for improvement.

Medea-Destiny:
The easiest way to do this is have the bookmarks app send the SLURL in chat to the person ordering a teleport. This would be a clickable map link in chat. I can add this easily for the next release

We could add a function to the remote that allows landmarks to be dropped into the remote to trigger simultaneous teleports where the user of the remote has RLV active, or if we reintroduce extended interface channel functionality (which I am in favour of) it would be pretty trivial to make a teleport hud that did this.

FunViking:
Results of the testing so far are as follows. The teleport information came up in my nearby and I clicked on it to teleport but then I got this message, you can only teleport 6 times in a minute.  Wait one minute and try again.   I waited over a minute and got the same message again.  collar is sending me somewhere different than where I force tp sub Serena (128,128,13) but it sent me to [20:57:31] Goannas Billabong, Serena (128, 128, 1): the collar is sending her to the correct tp location every time,but sending me to different places second time I went to Tropical (59,220,25) but I sent her to Tropical Lodge/170.000000/94.000000/27.000000.  The instructions you sent with the new script worked just as you had it on the notecard.

cruelraremaster:
Seems to be fixed with above PR

PR #716, opened 2021-12-17
Update oc_bookmarks.lsl

Medea-Destiny:
Added chat message to command issuer in TeleportTo() function that reformats into a clickable SLURL. This means that when the collar user teleports the collar wearer to a location in bookmarks, they will receive a chatted message informing them of where they sent the sub that is clickable to allow them to teleport directly to where they sent the sub.

This addresses issue #715

### DIFF
--- a/src/collar/oc_bookmarks.lsl
+++ b/src/collar/oc_bookmarks.lsl
@@ -1,10 +1,14 @@
 // oc_bookmarks
 // This file is part of OpenCollar.
-// Copyright (c) 2008 - 2017 Satomi Ahn, Nandana Singh, Wendy Starfall,  
+// Copyright (c) 2008 - 2021 Satomi Ahn, Nandana Singh, Wendy Starfall,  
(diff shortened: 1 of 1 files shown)


FunViking:
For the second day in a row when I try to TP sub I get the SL message stating unable to teleport more than 6 times a minute.  I have not tried to teleport more than 6 times in a minute.

Medea-Destiny:
> For the second day in a row when I try to TP sub I get the SL message stating unable to teleport more than 6 times a minute. I have not tried to teleport more than 6 times in a minute.

Just for the record, this is a (bugged) viewer message prompted by a TP failure. It may be caused by attempting to teleport to a sim that is offline or no longer exists, or might happen occasionally due to any number of other TP bugs -- it basically means the viewer was unable to contact the target sim. This isn't an issue with the script, which just generates a SLURL from the bookmark.

Pingout (approved):
Tested and approved by Omnia and Ping

Omnia2021 (commented):
Great feature, provides a fast travel option outside the usual route, works as designed.

PR #718, opened 2021-12-31
Addon Timeout Removal correction

KristenMynx:
The messages regarding removal of Addons did not correctly report the removals, because the list delete would remove 1 too many elements from the list.

### DIFF
--- a/src/collar/oc_addons.lsl
+++ b/src/collar/oc_addons.lsl
@@ -287,7 +287,7 @@ state active
                     }
                     SayToAddonX((key)llList2String(g_lAddons,i), "dc", 0, "", llGetOwner());
                     llMessageLinked(LINK_SET, NOTIFY, "0Addon: "+llList2String(g_lAddons,i+1)+" has been removed because it has not been seen for 5 minutes or more.", g_kWearer);
-                    g_lAddons = llDeleteSubList(g_lAddons, i, i+5);
+                    g_lAddons = llDeleteSubList(g_lAddons, i, i+4);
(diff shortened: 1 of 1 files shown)


KristenMynx:
I may have targeted the wrong branch, sorry, please correct as needed.

Medea-Destiny (commented):
Code change makes sense -- looping through blocks of 5 values in the list but deleting 6 values from the list to remove entries is obviously a mistake. 

To fulfil testing requirements we need someone to try this in world. Testing team, I suggest try to connect multiple add-ons to the collar, detach one of them and let be removed by timeout, and then check the others are still listed correctly (for example removing another after the first timeout occurs, and checking that one times out correctly, with the timeout message giving the correct name for the add on).

Pingout:
> Code change makes sense -- looping through blocks of 5 values in the list but deleting 6 values from the list to remove entries is obviously a mistake.
> 
> To fulfil testing requirements we need someone to try this in world. Testing team, I suggest try to connect multiple add-ons to the collar, detach one of them and let be removed by timeout, and then check the others are still listed correctly (for example removing another after the first timeout occurs, and checking that one times out correctly, with the timeout message giving the correct name for the add on).

We tested this script per your suggested method, and had no problem. We detached arm cuff, and checked the menu to be sure nothing else was removed. After the 5 minute timeout period, the arm cuff button was removed. The timeout was reported correctly. We also tried using the Disconnect button in the Addons Arm Cuff menu. The button was removed right away, and the cuff tried to reconnect (because it was still attached). The addon reconnected within moments, and the button reappeared in the Addons menu.

PR #727, opened 2022-01-25
moving settings editor to it's own script

Yosty7B3:


Yosty7B3:
Moving the editor into it's own script for now so it can be removed from oc_states.

(part of the steps for the overhaul of the bootup process to make it faster)

PR #728, opened 2022-01-25
moving settings editor to it's own script

Yosty7B3:
Moving the editor into it's own script for now so it can be removed from oc_states. 

(part of the steps for the overhaul of the bootup process to make it faster #732)

### DIFF
--- a/src/collar/settings_editor.lsl
+++ b/src/collar/settings_editor.lsl
@@ -1,0 +1,297 @@
+/*
+This file is a part of OpenCollar.
+Copyright ©2020
+: Contributors :
+Aria (Tashia Redrose)
+    *August 2020        -       *Created oc_states
+                                *Due to significant issues with original implementation, States has been turned into a anti-crash
+                                script instead of a script state manager.
+                                *Repurpose oc_states to be anti-crash and a interactive settings editor.
+Medea (Medea Destiny)
+    *July 2021          -       *See issue #587: Added warning when script resets folders script that user should consider cleaning
+                                up their #RLV 
+    Sept 2021           -       Tighten timings and number of passes on reboot process and reduced sleep padding.                               
+                                            
+    
+    
+et al.
+Licensed under the GPLv2. See LICENSE for full details.
+https://github.com/OpenCollarTeam/OpenCollar
+*/
+
+
+//integer CMD_ZERO = 0;
+integer CMD_OWNER = 500;
+//integer CMD_TRUSTED = 501;
+//integer CMD_GROUP = 502;
+//integer CMD_WEARER = 503;
+integer CMD_EVERYONE = 504;
+//integer CMD_BLOCKED = 598; // <--- Used in auth_request, will not return on a CMD_ZERO
+//integer CMD_RLV_RELAY = 507;
+//integer CMD_SAFEWORD = 510;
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
The settings edit functionality is currently provided by the oc_states script. There's no good logical reason to have it in the oc_states script other than that it has plenty of room in it. The logical place for this function is in the settings script, but we need the memory headroom in oc_settings. 

@Yosty7B3 's proposal puts this in a separate script. Yosty, can you outline the reasoning for this rather than just leaving it in oc_states please?

Yosty7B3:
I initially removed it for my own sanity to make sense of what was happening in the states script.  
But it also seemed to me for anyone who would look for it would never think to look in states to find it.

I realize oc_settings is already pretty full. But that would really be the most obvious place to put it. 
Keeping it in any other script would really mean that that script means the settings are being copies over.
In oc_settings the editor could access the settings directly.

But if oc_settings really isn't a good place then keeping it in it's script is probably the most logical solution. 
Maybe we can even move some other parts like the remote settings into that same script to clean up the main setting script even more.

Medea-Destiny:
Two things. First is that removing the settingsedit function at this point may be good for your sanity in helping understand what oc_states is actually trying to do, but it's bad for our sanity in helping the rest of us understand your changes to oc_states per #732 :D Because it's a great big chunk of code gone from the oc_states script it means that everything in the script is moved around and that makes it harder to focus on the actual changes that #732 is implementing. I'd like to see settingsedit remain in oc_states while we focus on your speedup changes.

Second thing is that it occurs to me that there has been a significant call to remove Weld functionality from the core scripts and make it app based. Support has had to deal with a fair number of people who've got themselves welded without really knowing what weld does, and having that extra step of going through an app would probably be a good idea. This is relevant because there's a fair bit of code in oc_settings that is part of the weld functionality, and if we move weld entirely out of the core scripts, it looks to me like fitting settingsedit into oc_settings, which I'm sure we all agree is the ideal place for it, suddenly becomes a realistic target. 

Taking both these thoughts into account, I suggest we leave settings editor in oc_states for the moment, and focus on your speedup changes and appifying Weld, and come back to this change after we're good with those two.

Yosty7B3:
I've reworked states after I removed editor, so if editor needs to get back in I have to add it in. I have no version of the new states with editor still in it.
I can put it back in though, shouldn't be a problem.

In #730 I did include the removal of editor from the original oc_states as a commit before the full rework of states, so you can check the changes between the old and new without the editor changes complicating the comparison.
https://github.com/OpenCollarTeam/OpenCollar/pull/732/commits/0ed8b40564e0c0d9e7d697d7f38f9323daa23838

There hasn't been much left of the old code though, so comparing it probably isn't the best way to understand what's going on.
I do hope it shouldn't be to hard to understand though :)

So should I add editor back in then?

Yosty7B3:
Weld in it's own script would be great. 
I did run settings with editor in it and had no error messages. 
So with weld removed it should just be enough to run safely.

SilkieSabra:
I just want to chime in that this seems like a great direction to me.  As a support person, it's painful to see the line of people trying to unweld after they went "oooh what does this do" and pushed a button in the main menu.  And it suits the OpenCollar ethos more to have weld as an optional app.

PR #732, opened 2022-01-25
Speedup boot process in oc_states

Yosty7B3:
Uses feedback from other scripts instead of arbitrary delays to trigger the steps in the boot sequence.

      ~~ boot states ~~
default: REBOOT reboot         
            ALIVE                   from oc_api & oc_core & oc_settings
startup: STARTUP     
            0 initialize            from oc_core
            LM_SETTING_REQUEST ALL  from oc_api
            "settings=sent"         from oc_settings
running: TIMEOUT_READY          

(part of the steps for the overhaul of the bootup process to make it faster. just like #730  & #728 )

PR #733, opened 2022-02-01
only reset when needed

Yosty7B3:
These changes should help make the script available quicker after a relog or reattach.

Request ALL settings will still be sent to make sure setting are going to be send and received after a STARTUP message.
Other then that, a reset should only be needed on CHANGED_OWNER.

( independent addition to #732 )

Simular changes can be made to oc_core as well (and probably other scripts), but waiting for feedback in this first.


kinchcomputers (approved):

Medea-Destiny:
Closing to transfer over to #866 - same code, just streamlining merge.

Issue #743, opened 2022-02-22
OpenCollar Addon Cuffs chains fail in certain situations


Pingout:
What version of OpenCollar are you using? Current version 8.2, OpenCollar Addon Cuffs version 2.0

What behavior did you expect? Chains should not disappear unexpectedly

What behavior did you see instead? Last night we were able to see a problem in the cuffs, where Leg chains would disappear when an Arms pose was selected.  The Legs pose stayed active, just without the chains.
Eventually we kinda piece back together the steps that caused the problem:
our original problem was the restrictions wouldn't clear between poses
in fact restrictions didn't clear at all, unless your owner used the Clear Chains button in the Addon menu
as we looked into the problem we found 2 lines in the script which were commented out with the double slash //
both had a pointer to CLEAR_ALL_CHAINS, a subroutine that only the owner could get to
Our thinking at the time of Version 2 Release was if we remove the comment marks, activating those two lines
the restrictions would clear between poses. What we noticed yesterday was not only do the restrictions clear, but the chains as well

I tried a fix for the problem by creating a CLEAR_POSE_RESTRICTION routine to use in place of the CLEAR_ALL_CHAINS
(see associated Pull Request)

KristenMynx:
PR #745 merged, closing

PR #745, opened 2022-02-22
Updated oc_cuff script

Pingout:
Reference Issue #743  

This fix allows RLV Restrictions to be placed on a per-pose basis, clearing each set of restrictions between one pose and the next, and also clearing restrictions on a *STOP* from the Addon Pose Menu


Medea-Destiny (approved):
Approval of code: Code is in style and contains no obvious problems. I'm not hugely familiar with cuff scripts however.

Medea-Destiny:
Changed base to features branch for new scripts.

PR #751, opened 2022-03-07
Change's to sounds for installer

StormedStormy:
added an additional sound in the installer for per update to sound a lot better rather than a few beeps, but to also make the user aware of prompts.


Medea-Destiny (approved):
Looks good to me.

SilkieSabra:
Since this was contributed by Stormy and Medea approved it i'm going to merge.  If anyone thinks that's too soon we can roll it back.

PR #755, opened 2022-03-17
Changes to Help/About

StormedStormy:
Added a reboot button so users can reboot there collars easier as some do not know about the prefix

### DIFF
--- a/src/collar/oc_core.lsl
+++ b/src/collar/oc_core.lsl
@@ -24,6 +24,8 @@ Medea (Medea Destiny)
                 -   Added confirmation messages when group or public access is toggled and fixed a typo
                 -   Efficiency pass, inlined majorminor(), docheckupdate() and docheckdevupdate().
                     Removed g_lTestReports, left over from alpha.a
+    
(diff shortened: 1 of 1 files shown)


Medea-Destiny (changes requested):
The code simply provides a button in the help menu to access the reboot function, as if the reboot chat command had been given. Code is fine, but there's a small style change to match the standard formatting that is needed:
```
} else if(sMsg == "Reboot")
                        {llMessageLinked(LINK_SET, iAuth, "Reboot", kAv);
```
...should read as
```
} else if(sMsg == "Reboot") {
                        llMessageLinked(LINK_SET, iAuth, "Reboot", kAv);
```
... note that OC style guide uses a braces format where the opening curly brace should be at the end of the conditional line, not on the line below it. Thanks!

StormedStormy:
> This would appear to be in error. There is still a { on line 688. If the intent was to pull it up then this change is incomplete. Does this actually compile?

it still worked regardless if its there but it it wants put a certain way then ill have to change it again

StormedStormy:
its been changed now and have gone ahead and removed it

Medea-Destiny (approved):
Code is now good.

PR #763, opened 2022-04-04
Update oc_anim.lsl

K9K8E:
Removed  CONTROL_ML_LBUTTON to allow left clicking while in mouselook.


Pingout:
This resolves a bug reported in Issue 747 where the collar was not allowing the wearer to click on objects while in mouselook

Issue #765, opened 2022-04-06
Problem with Synclock in OpenCollar Addon Cuffs


Pingout:
What version of OpenCollar are you using? Collar 8.2, Cuffs Version 2.1 Beta 2

Synclock is an option in the OpenCollar Addon Cuffs script. Synclock is supposed to make the cuff locks synchronize with the collar locking/unlocking procedure. With Synclock active, the collar lock will operate the cuff locks. Cuffs locking should work using Lock or Synclock, without interfering with restrictions set by notecard per pose

In some circumstances the Cuffs do not lock correctly using Synclock. For example you need to cycle the collar lock before the cuff and collar locks synchronize,  and also the script will unlock the cuffs on a pose change which applies restrictions. 

If you are wearing cuffs and have Synclock checked but you do not lock the cuffs first, you can watch the restrictions console as the cuffs remove the detach=n. However if the cuffs are locked first, then Synclock is applied, the pose restrictions are applied and removed without error. This is counter-intuitive, and most users will not think they need to lock the cuffs before applying Synclock.

A possible solution would be to remove the Synclock feature. A case could be made that this feature is really not necessary. Alternatively we will need to dig into the script to see why the Synclock is causing this problem.

Pingout:
Thanks and a tip of the hat to Katie (k9k8e) for discovering this bug.

MuttMode:
Opinion but logical, I hope: Cuffs are assumed to be a separate 'device' to the collar so having their own, entirely independent lock seems the most sensible option. (And FAR less work than tracking down whatever's causing the bug)
People usually wear their collar permanently but cuffs are often applied 'as needed' and removed after a scene so having their own lock would be preferable in that sense, too.

KristenMynx:
closing since the pull request #768 was closed and the change was implemented in #771

mistressohm:
Sync-Lock is a toggle-able option, allowing someone to 'permanently' wear
cuffs along with the collar.   If it's not desired, uncheck it in the Cuff
options.

As for why the bug is occurring, it appears that Sync-lock only 'watches'
the collar Lock state after it is set, and doesn't actually check the
collar Lock state (and enforce it with the Cuffs) when first engaged.

So, here's how I would propose the app should work:

*Case: Trusted/Wearer *

Action: SyncLock is checked in Cuff options.

SyncLock app checks the Lock state of the collar.
if unlocked, it sets the cuffs to unlocked and then goes into 'watch' mode
[...]

PR #768, opened 2022-04-11
Update oc_cuffs to remove SyncLock Function

Pingout:
Please reference Issue #765


Medea-Destiny (changes requested):
As discussed, remove commented code rather than leaving it all commented. 

Style issue: open curly braces should be on the end of the if... line, in numerous places they are on a line on their own. While I find that much easier to read myself, it's not the approved OC style.

There may be issues with if(sStr=="settings=sent"){ section , but that will be easier to see with the comments removed. Did I mention I don't like this brace style?

KristenMynx:
I believe this is now stale, handled by #771 should we close this?

Pingout:
> I believe this is now stale, handled by #771 should we close this?

That is correct. #771 has been merged. Closing this Issue

PR #771, opened 2022-04-13
Update oc_cuff.lsl

Pingout:
Remove commented code, fix style issue re: open curly braces should be on the end of the if... line

Reference PR #768


Medea-Destiny (changes requested):
There are double braces in a couple of places where an if clause has been removed, see line 623 and 667. Also a few instances of open curly brace on its own line still. I don't see anything here that will impact running the code, but those should be cleaned up for style and clarity.

Issue #778, opened 2022-05-07
mouse look cross sim/teleport


tayaphidoux:
What version of OpenCollar are you using? 
8.2.1000

What behavior did you expect?
to stay in mouse look 

What behavior did you see instead?
the viewer canceled mouse look on teleport and the mouse got stuck free

What steps does someone need to take to reproduce the problem?
set mouse look restriction i.e. camdistmax:0

i have this fixed on my own custom add-ons with a simple if region changed set unset then reset cam distance.

Medea-Destiny:
Unable to reproduce in Firestorm -- which viewers do this? It should be reported as a viewer bug.
@tayaphidoux or anyone else who can reproduce this issue, can you please also try out for other values of maxcamdist, and also mincamdist?

tayaphidoux:
as i mentioned it happens on viewers that by default force you out of mouse look when you teleport, the one i know for sure does it is singularity i can test others but i do remember having this issue before on firestorm.

tayaphidoux:
just checked fire storm it does do it but it don't actually force mouse when max distance is 0 look to begin with you can still toggle between mouse look and not in mouse look

tayaphidoux:
it does not break in kokua.

tayaphidoux:
@Medea-Destiny  you are welcome to come up with a different solution if you think that re en stating the camera distance when ever the sim changes is to much, and have taken my solution cross sim carting before and teleporting around multiple shops with my owner.

tayaphidoux:
catznip don't actually have a linux client yet, and light speed i don't trust, coolvlviewer does not accept my password its a bug that used to affect the official viewer, and since the official viewer don't have rlv by default there is no reason to test it, i can only test viewers with linux support.

tayaphidoux:
@Medea-Destiny 
    "Mouselook"     , 7 , "camdistmax:0"                            // 61  31 <- part of oc_rlvsuit.lsl that refrences mouse look.

below also part of oc_rlvsuit.lsl its function is to set mouselook aka (@camdistmax:0)
```
else if ((sCommand == "setcam_avdistmax" || sCommand == "setcam_avdistmin") && g_bForceMouselook) {
        sCommand = "camdistmax:0";
    } else if (sCommand == "camdistmax:0"){
        if (bEnable) g_bForceMouselook = TRUE;
        else g_bForceMouselook = FALSE;
```
my solution which calls the code above to toggle mouse look and resolve this bug.
```
[...]

PR #780, opened 2022-05-07
fix camdist 0 bug

tayaphidoux:
this should fix a bug on viewers that kick you out of mouse look on teleport or sim crossing.

### DIFF
--- a/src/collar/oc_rlvsuite.lsl
+++ b/src/collar/oc_rlvsuite.lsl
@@ -595,7 +595,7 @@ state active
     }
     
     change(integer changed){
-        if(changed && CHANGE_REGION){
+        if(changed && CHANGED_REGION){
             if( g_bForceMouselook == TRUE){
                 FormatCommand("camdistmax:0",FALSE);
                 FormatCommand("camdistmax:0",TRUE);
--- a/src/collar/oc_rlvsuite.lsl
+++ b/src/collar/oc_rlvsuite.lsl
@@ -597,7 +597,12 @@ state active
     changed(integer change){
         if(change && CHANGED_REGION){
             if( g_bForceMouselook == TRUE){
+                // Disable Mouse look
                 FormatCommand("camdistmax:0",FALSE);
+                // force the user out of mouse look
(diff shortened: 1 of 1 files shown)


tayaphidoux:
https://github.com/OpenCollarTeam/OpenCollar/issues/778 - i describe the issue this code fixes here.

Medea-Destiny:
@tayaphidoux The code seems fine here, however I'm still not clear this is a bug we actually _should_ be making a workaround for. As I understand it's something that so far has only been reported with Singularity? Shouldn't this then be reported on the Singularity JIRA for them to fix?

tayaphidoux:
@Medea-Destiny then try it out for your self put your self in forced mouse look and try to teleport, on a viewer other than firestorm,just because firestorm likes to violate the shared experience by being the only viewer to not properly engage mouse look based on distance from avatar, does not mean that its not a bug that you can't turn and move when your camera is locked into mouse look like it should be.

KristenMynx (approved):
The changes look fine.   I don't see any harm sending a couple more viewer messages on region change.  Yes it probably should be a viewer fix, not an OC fix, but it doesn't look expensive to me.

tayaphidoux:
its a work around i am not blaming oc for causing it its a miss comparability between rlv calls and viewers default functionality as defined by linden labs, so most viewers will comply unless they actually test the features to ensure compatibility, in the mean time this was what i came up with.

> The changes look fine. I don't see any harm sending a couple more viewer messages on region change. Yes it probably should be a viewer fix, not an OC fix, but it doesn't look expensive to me.

tayaphidoux:
@KristenMynx should i point this towards q1?

KristenMynx:
2022-q1-features yes.  Also lets see if there are any more comments before merging.

tayaphidoux:
i guess its a good thing this is not merged yet as there appears to be a bug created with this, its not maintaining mouse look on login with my fix.

tayaphidoux:
i am closing this for now because this causes an issue with mouse look and i need to figure out why

Issue #779, opened 2022-05-07
oc_outfits has changing paths for outfits under #RLV


AlexNumberFour:
What version of OpenCollar are you using? 8.2.1000

What behavior did you expect? 

I expected to be able to set up my outfits under #RLV/.outfits 

What behavior did you see instead?

Sometimes oc_outfits  changes it's mind and I have to rename #RLV/.outfits to #RLV/outfits for it to see my outfits. The dialog window simply doesn't show my outfits so they can be selected otherwise. 

I've had to change back and forth between #RLV/.outfits and #RLV/outfits several times now.

What steps does someone need to take to reproduce the problem?

1.  I have not found a way to reliably reproduce the problem other than to just use it regularly. 

As best I can tell, looking at the code, 

// from: oc_outfits

// looks like this has a problem setting g_iRLVa sometimes ... 
// however, I have no idea what is sending these messages -- that's up to someone with more knowledge than I have to figure out.
    listen(integer iChan, string sName, key kID, string sMsg)
    {
        if(llSubStringIndex(sMsg, "RLVa")==-1){
            g_iRLVa=FALSE;
        }else g_iRLVa=TRUE;
    }

// this causes the code to return the wrong value sometimes based on g_iRLVa
string GetOutfitSystem(integer iCorePath){
    if(g_iRLVa){
        if(iCorePath)return ".outfits/.core";
        else return ".outfits";
    }else{
        if(iCorePath)return "outfits/core";
        else return "outfits";
    }
}

I've attached a copy of the oc_collar file that is in my collar

[oc_collar.txt](https://github.com/OpenCollarTeam/OpenCollar/files/8644252/oc_collar.txt)

tayaphidoux:
it should be using ~outfits instead of .just plain outfits

AlexNumberFour:
I'm using the latest version of 64 bit firestorm, 6.5.3.65658. It's the only viewer I use for SL. 

> @versionnew=0
versionnew: RestrainedLove viewer v3.4.3 (RLVa 2.4.2)

I always have RLV enabled.

tayaphidoux:
given that info it should only do .outfits and .outfits/.core seems the code needs to be checked out its possible some one changed somthing or its just not coded flexible enough

tayaphidoux:
it checks the rlv version here 

lines 348-350 as of this post

 listen(integer iChan, string sName, key kID, string sMsg)
    {
[...]

PR #781, opened 2022-05-07
fix Outfits/Core non RLVA Compatability and RLVa Check

tayaphidoux:
fix the compatability of nonRLVA viewers without exposing the folder to normal usage.

also applies fix for https://github.com/OpenCollarTeam/OpenCollar/issues/779

### DIFF
--- a/src/Apps/oc_outfits.lsl
+++ b/src/Apps/oc_outfits.lsl
@@ -313,7 +313,7 @@ integer ALIVE = -55;
 integer READY = -56;
 integer STARTUP = -57;
 
-integer g_iRLVa = 0;
+integer g_iRLVa = FALSE;
 
 string RLVA_APPEND=".";
 
@@ -367,13 +367,12 @@ state active
     {
         if(llGetStartParameter()!=0)llResetScript();
         g_kWearer = llGetOwner();
-        //llMessageLinked(LINK_SET, LM_SETTING_REQUEST, "global_locked","");
-        llMessageLinked(LINK_SET, LM_SETTING_REQUEST, "ALL","");
+        llMessageLinked(LINK_SET, LM_SETTING_REQUEST, "global_locked","");
     }
     link_message(integer iSender,integer iNum,string sStr,key kID){
         if(iNum >= CMD_OWNER && iNum <= CMD_EVERYONE) UserCommand(iNum, sStr, kID);
-        else if(iNum == RLV_VERSION || iNum == RLVA_VERSION ){
-               if(~llSubStringIndex(sStr, "RLVa")){
+        else if(iNum == RLVA_VERSION ){
+               if(sStr == "1"){
                   g_iRLVa=TRUE;
                }else g_iRLVa=FALSE;
         }
--- a/src/Apps/oc_outfits.lsl
+++ b/src/Apps/oc_outfits.lsl
@@ -59,8 +59,13 @@ integer MENUNAME_RESPONSE = 3001;
 integer OUTFITS_ADD = -999901;
 integer OUTFITS_REM = -999902;
 
-//integer RLV_CMD = 6000;
-//integer RLV_REFRESH = 6001;//RLV plugins should reinstate their restrictions upon receiving this message.
+integer RLV_CMD = 6000;
+integer RLV_REFRESH = 6001;//RLV plugins should reinstate their restrictions upon receiving this message.
+integer DO_RLV_REFRESH = 26001;//RLV plugins should reinstate their restrictions upon receiving this message.
+integer RLV_CLEAR = 6002;//RLV plugins should clear their restriction lists upon receiving this message.
+integer RLV_VERSION = 6003; //RLV Plugins can recieve the used RLV viewer version upon receiving this message..
+integer RLVA_VERSION = 6004; //RLV Plugins can recieve the used RLVa viewer version upon receiving this message..
+integer RLV_CMD_OVERRIDE=6010; //RLV Plugins can send one-shot (force) commands with a list of restrictions to temporarily lift if required to ensure that the one-shot commands can be executed
 
 //integer RLV_OFF = 6100; // send to inform plugins that RLV is disabled now, no message or key needed
 //integer RLV_ON = 6101; // send to inform plugins that RLV is enabled now, no message or key needed
--- a/src/Apps/oc_outfits.lsl
+++ b/src/Apps/oc_outfits.lsl
@@ -372,7 +372,7 @@ state active
     link_message(integer iSender,integer iNum,string sStr,key kID){
         if(iNum >= CMD_OWNER && iNum <= CMD_EVERYONE) UserCommand(iNum, sStr, kID);
         else if(iNum == RLVA_VERSION ){
-               if(sStr == "1"){
+               if(sStr != "0"){
                   g_iRLVa=TRUE;
                }else g_iRLVa=FALSE;
         }
--- a/src/Apps/oc_outfits.lsl
+++ b/src/Apps/oc_outfits.lsl
@@ -340,14 +340,6 @@ default
     }
     state_entry(){
         llMessageLinked(LINK_SET, ALIVE, llGetScriptName(),"");
-        llListen(999988, "", llGetOwner(), "");
-        llOwnerSay("@versionnew=999988");
-    }
-    listen(integer iChan, string sName, key kID, string sMsg)
-    {
-        if(~llSubStringIndex(sMsg, "RLVa")){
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
Switching from @version to @versionnnew is good practice, but I don't see the purpose of your change to the version checking code -- can you provide a case in which your version will provide a different answer?

We already have a version check being done in oc_rlvsys -- it seems silly to be doing it twice, and I'd suggest it would be better for the folders script to determine if we're using RLV or RLVa from the RLVa version message from oc_rlvsys.

tayaphidoux:
the @versionnew is just because its what is supposed to be used it was using @version the rest ~llSubStringIndex() instead of llSubStringIndex() == -1 may be more accurate and prevent the swapping issue mentioned in the issue, using ~ in the outfit folder instead of . works on viewers that don't use RLVa with the same functionality that it does on RLVa viewers it hides the folder from #Folders where as just removing  the . exposes it.

> Switching from @Version to @versionnnew is good practice, but I don't see the purpose of your change to the version checking code -- can you provide a case in which your version will provide a different answer?
> 
> We already have a version check being done in oc_rlvsys -- it seems silly to be doing it twice, and I'd suggest it would be better for the folders script to determine if we're using RLV or RLVa from the RLVa version message from oc_rlvsys.

tayaphidoux:
this will need to be tested now to ensure that fix works took me some digging to find out how to confirm the RLV version suing oc_rlvsys.lsl and for all intents and purposes it could not work because i have done some thing wrong i am still in the learning stage when it comes to the cross script communication

KristenMynx:
I'm willing to review the code and test in-world, however the discussion under #779 (debugging, discussions, alternatives) made my mind go TL;DR.  A direct discussion (in-world or discord) (understanding which viewers involved as well) would probably be easier for me to check this out.

tayaphidoux:
> I'm willing to review the code and test in-world, however the discussion under #779 (debugging, discussions, alternatives) made my mind go TL;DR. A direct discussion (in-world or discord) (understanding which viewers involved as well) would probably be easier for me to check this out.

ok simple rule if you go into mouse look and teleport and you are kicked out of it that viewer will be affected.

the only exception is how the viewer handles mouse look entrance, most viewers if you zoom(scroll wheel) in will auto hook mouse look at a specific distance, i think that's an opt in for firestorm.

KristenMynx:
Ummm... @tayaphidoux this pull request doesn't reference mouse look.   It has to do with (dot) and (tilde) folder prefixing

tayaphidoux:
> Ummm... @tayaphidoux this pull request doesn't reference mouse look. It has to do with (dot) and (tilde) folder prefixing

that is what i am trying to make clear and it does more than that i have been working to try and solve the folder switching problem mentioned in #779 am just trying to make it clear that the pointer in mouse look issue is wrong.

Medea-Destiny:
@tayaphidoux I'm not sure about the change to using ~ -- while I see the logic behind it, it's going to break people's current set-ups.

tayaphidoux:
only if they are using a RLV viewer instead of an RLVa viewer however the change might be apprecated because then the folder will be properly hidden and work like it does in RLVa with the (dot)

Medea-Destiny:
@tayaphidoux Yes, I get the logic behind it, and really  it makes sense to me. However I'd like to hear more opinions from our support team though, because this is a change that would require RLV (not RLVa) users to rename folders for functionality to keep working.

KristenMynx:
@Medea-Destiny  has a point.  If this should then be implemented, it should be enabled with a settings toggle.   Possibly not even checking RLV/RLVa version, but relying on the setting, with enough docs/help to explain what they have to change.

tayaphidoux:
@KristenMynx I have no issues with going this route it will just take more extensive work on the script to make the menu option, which is a bit above my head at the moment since i have not even gotten into configuring menus for oc on that level.

> @Medea-Destiny has a point. If this should then be implemented, it should be enabled with a settings toggle. Possibly not even checking RLV/RLVa version, but relying on the setting, with enough docs/help to explain what they have to change.

KristenMynx:
@tayaphidoux  I've reviewed this more carefully and my comment about the additional menu item doesn't make sense.   

I see its a race condition that you're resolving by using the RLV_VERSION and RLVA_VERSION link messages to detect rlv/rlva instead of having oc_outfits ask for a version directly.   This does make sense.

One question remains in my mind:  why did you change 
        llMessageLinked(LINK_SET, LM_SETTING_REQUEST, "global_locked","");
to
        llMessageLinked(LINK_SET, LM_SETTING_REQUEST, "ALL","");

This seems to be irrelevant to the change, and would cause excessive link message chatter.   During startup, another script (core?) calls for all the settings to be reported, so oc_outfits should see the settings come through that.

tayaphidoux:
because i didn't know if global_locked would send the RLV and RLVA check, and the one way i saw that it gets reported is in the all function, also i wanted a check call from the script just in case it misses it on startup.
>

KristenMynx:
RLV_VERSION and RLVA_VERSION is not stored in settings.  rlvsys manages that.   For some reason a request settings 'all' will cause rlvsys to send that information, but I'm not sure why it should need to.   The startup cycle of the collar (down in the timer) should be doing this.   It will fail if you reset/recompile the one script, but that is no reason to ask for the entire set of settings to be pulled.  I will test without the 'all', and check if bad/missing version strings are ever seen by oc_outfits.

tayaphidoux:
@KristenMynx that is why i left the other call commented but in the script, if it works and my change is redundant that is fine we are a team, i just have a habit when i code trying to consider how it can be broken and how to prevent it, and in this case it seems possible that lag could cause it to fail generating the issue like what the op described in #779

KristenMynx:
The fix won't work anyway.  RLV_VERSION and RLVA_VERSION return only numbers.  RLVa is not part of the value.

[11:39] Kristen's @ - Silky Ribbons Open Collar: rlvversion: 204
[11:39] Kristen's @ - Silky Ribbons Open Collar: rlvaversion: 304
[11:39] Kristen's @ - Silky Ribbons Open Collar: rlvversion: 204

rlvsys checks the version and sets integer values:
g_iRlvVersion
g_iRlvaVersion

g_iRlvaVersion will be 0 if RLVa


So, to check for RLVa, the proper check is to look for RLVA_VERSION and check for "0".   Checking for RLV_VERSION is therefore irrelevant.

tayaphidoux:
> The fix won't work anyway. RLV_VERSION and RLVA_VERSION return only numbers. "RLVa" is not part of the value.
> 
> [11:39] Kristen's @ - Silky Ribbons Open Collar: rlvversion: 204 [11:39] Kristen's @ - Silky Ribbons Open Collar: rlvaversion: 304 [11:39] Kristen's @ - Silky Ribbons Open Collar: rlvversion: 204
> 
> rlvsys checks the version and sets integer values: g_iRlvVersion g_iRlvaVersion
> 
> g_iRlvaVersion will be 0 if RLVa
> 
> So, to check for RLVa, the proper check is to look for RLVA_VERSION and check for not = "0". Checking for RLV_VERSION is therefore irrelevant.

i kinda figured as much however because i was not fully sure on how the code was checking in rlvsys and if it populated both or just one i did that in case, like i said in the post above redundancy does happen some times with my code as i try to contemplate as many possible breaks and provide solutions, once its  working then i go back and clean it up to make sure i don't have any truly unnecessary redundancy

tayaphidoux:
@KristenMynx  changes applied and because we are only checking for RLVa we have to assume false if it fails.

KristenMynx:
@tayaphidoux  The test against RLVA_VERSION sStr it needs to be 
...
 if(sStr != "0"){
...
The number that comes through is an integer version number.  In my testing it came through as "304" in the latest production firestorm.   The rest is fine.

tayaphidoux:
@KristenMynx you are correct i just opened llOwnerSay on the section that spams the settings ping and searched for 6003 and 6004 and relayed the message so i could see it mine says 104 there is also a part where it sets FALSE in the code i didn't read more and assumed the logical answer was TRUE else but just seems to only spit out the version number.

tayaphidoux:
tested this in world finally didn't cause any issues and it works correctly for "RrestrainedLove Viewers"

SilkieSabra:
Can we get an inworld copy of this script also for testing?  What would be the proving test for this?

tayaphidoux:
@SilkieSabra  dunno if you know already but i did send it out when you asked for the other script to be compiled and given out, i have also been actively testing it.

Pingout (approved):
I tested this inworld for a two days, in various outfits. Everything works fine
@SilkieSabra this can be merged

Issue #775, opened 2022-05-01
Rogue Chat Command !!


SilkieSabra:
it seems that if you type [prefix] - and either leave the value blank or insert anything that's not an existing folder name, the chat command will strip everything on your avatar, including items that are not in any #RLV folder, and including items in .outfits -- failing only to strip items in a locked .core folder.

StormedStormy:
> it seems that if you type [prefix] - and either leave the value blank or insert anything that's not an existing folder name, the chat command will strip everything on your avatar, including items that are not in any #RLV folder, and including items in .outfits -- failing only to strip items in a locked .core folder.

Confirmed issue - StormedStormy but have also found where the culprit is in the folders script too

Medea-Destiny:
Looks like a lack of sanity checking, should be an easy fix.

Medea-Destiny:
Fix done in world. I have cleaned up the string handling so that it will not attempt to search for the characters that are intended to be commands when the string is truncated -- i.e (prefix)- would search for - because the search term was found with an llGetSubString command that assumes there are characters after the command to work. Replacing with llDeleteSubString ensures we never pass the command to the search. Empty strings are no longer searched for, and searches that produce no result no longer attempt to attach/remove "". Instead the command issuer is given an warning that nothing was found with the search term.

I'll pass this script along for testing before doing the pull.

KristenMynx:
I've done a code review.   All changes look like they correct the problem.   I'll advise when I test.

KristenMynx:
After testing I found that   [prefix] -X   (single character folder specified) results in 'You have to supply a string to search for to use add/remove/wear commands.'

Proposed fix:
Remove line 203
Add the following in its place:
        string sChangevalue = llStringTrim(llDeleteSubString(sStr, 0, 0), STRING_TRIM);
        // double check for double && ++
        if(llListFindList(["&","+"],[llGetSubString(sChangevalue,0,0)]) > -1) { 
            sChangevalue = llStringTrim(llDeleteSubString(sChangevalue, 0, 0), STRING_TRIM);
        }

I'd also suggest changing the messages:

'You must supply a folder name to search for to use add/remove/wear commands'
to
'You must supply a folder name to when using add/remove/wear folder commands'

PR #783, opened 2022-05-12
Update oc_folders.lsl

Medea-Destiny:
Fix for issue #775.  Cleaned up the string handling so that it will not attempt to search for the characters that are intended to be commands when the string is truncated -- i.e (prefix)- would search for - because the search term was found with an llGetSubString command that assumes there are characters after the command to work. Replacing with llDeleteSubString ensures we never pass the command to the search. Empty strings are no longer searched for, and searches that produce no result no longer attempt to attach/remove "". Instead the command issuer is given an warning that nothing was found with the search term.

### DIFF
--- a/src/collar/oc_folders.lsl
+++ b/src/collar/oc_folders.lsl
@@ -1,20 +1,19 @@
 /*
 This file is a part of OpenCollar.
 Copyright 2020
-
 : Contributors :
-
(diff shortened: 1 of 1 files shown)


tayaphidoux (approved):
does not break the collar unfortunately i don't know how to test this bug based on the descriptions, so i assume it does work.

PR #787, opened 2022-05-17
Create Cuff Themes

KristenMynx:
oc_cuff_themes is an optional script which eavesdrops on collar addon messages watching for theme change requests.  The normal oc_themes must also be in the collar.  If a matching theme notecard is found in the cuff, it is applied. 

oc_cuff_themes needs to be in each cuff, and the theme notecards in each as well.  The .theme notecards have the same format as the collar. Each cuff theme needs to have the same name as its corresponding theme in the collar.   Cuffs don't require all the themes that the collar has.  Theme change requests that have no corresponding notecard are silently ignored.   The notecards in the cuffs specify their own faces, textures and colors.   The cuff notecards contents don't need to be (nor should they be) the same as the collar notecards, since usually the cuff faces and links are different.

### DIFF
--- a/src/cuffs/oc_cuff.lsl
+++ b/src/cuffs/oc_cuff.lsl
@@ -167,6 +167,10 @@ Menu(key kID, integer iAuth) {
         if(!g_iHidden)
             lButtons+=["Pose"];
         else sPrompt +="\nPoses not available while the Collar is hidden";
+	 if(llGetInventoryType("oc_cuff_themes")==INVENTORY_SCRIPT){
+        if(!g_iHidden)
+            lButtons+=["New Theme"];
+        }        
     }
 
     if(iAuth == CMD_OWNER) {
--- a/src/cuffs/oc_cuff.lsl
+++ b/src/cuffs/oc_cuff.lsl
@@ -167,7 +167,8 @@ Menu(key kID, integer iAuth) {
         if(!g_iHidden)
             lButtons+=["Pose"];
         else sPrompt +="\nPoses not available while the Collar is hidden";
-	 if(llGetInventoryType("oc_cuff_themes")==INVENTORY_SCRIPT){
+	 if(llGetInventoryType("oc_cuff_themes")==INVENTORY_SCRIPT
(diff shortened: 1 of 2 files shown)


Pingout:
So yes! Very cool, aside from the verbosity when changing the theme on a collar and nine cuff set. 
   I do see one potential problem, that being the .theme notecard creation.  Experienced OC Themes users will have no trouble at all in creating a set of .theme notecards, but the inexperienced user might try clicking the [New Theme] button in the collar menu, and then get stumped when the theme from the collar does not work well on the cuff, and there's no [New Theme] routine to automatically generate the notecards they need for the cuffs. As I said, this is only a potential problem, and the instructions for making the cuff .theme notecard are already in the Collar Complete User's Guide.
  I will look at revising the Cuffs Builder Guide

KristenMynx:
In my collar/cuff sets, I'm making extensive use of themes (44 notecards affecting many aspects of the collar look, with I believe 11 in the cuffs) Contact me in-world and I'll send a set to you.

Pingout:
"The more I think about it the more I like having separate .theme notecards in the cuff, nothing can get mixed up. [New Theme] overhead experienced by the user should be minimized by the fact that most cuffs have the same link and face layout. New Theme once and put the notecard in all cuffs. Or perhaps only in one cuff (flagged by the config notecard) and transmitted to all other cuffs. oc_cuff_theme could run with 1 of 2 modes, independent or slaved. Independent means use theme notecards, and transmit the data via chat to the other cuffs that might be listening. Slaved, meaning it only listens for theme data on a channel. Possible issues with lag and chat packet dropping could be an issue."

Seperate .theme notecards in each cuff works well. I edited the Cuff Builder Guide to include Theme support. I think with the necessary caviats most, if not all builders will be able to utilize this functionality.  We may still get the unexperienced user who breaks their cuffs by keeping the Collar Rotation and Size information in the cuff .theme notecard.
[OC Cuffs Builder Guide May 2022.txt](https://github.com/OpenCollarTeam/OpenCollar/files/8748095/OC.Cuffs.Builder.Guide.May.2022.txt)

KristenMynx:
I added New Theme support to this Pull request

Pingout (approved):
Awesome, the themes work great. We incorporated a handful of themes into the OpenCollar Addon Cuffs, so people can see how they work.
@SilkieSabra you can merge this script at your convienience

PR #791, opened 2022-05-19
"offline" commands for addons corrupt addon list

KristenMynx:
The section of code which removes an addon from the addon list (g_lAddons) when and 'offline' chat message is received removes one too many list elements, thereby corrupting the list.  The stride is 5, but 6 elements (ipos, ipos+5 = 6) are removed.   This problem is similar to #718

### DIFF
--- a/src/collar/oc_addons.lsl
+++ b/src/collar/oc_addons.lsl
@@ -369,7 +369,7 @@ state active
                         {
                             llOwnerSay("Pre-Addon Deletion List: "+llDumpList2String(g_lAddons,"~"));
                         }
-                        g_lAddons = llDeleteSubList(g_lAddons, iPos, iPos+5);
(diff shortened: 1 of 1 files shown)


Medea-Destiny (approved):
Change looks good to me. Pending in-world tester approval. @Pingout can you handle this? Needs to check everything continues to work as expected with no unexpected timeout messages if you have two or more add-ons active and let one of them timeout.

Pingout:
> Change looks good to me. Pending in-world tester approval. @Pingout can you handle this? Needs to check everything continues to work as expected with no unexpected timeout messages if you have two or more add-ons active and one of them sends an offline message.

I tested this with a couple of different addons, so far-so good!

Pingout (approved):
No problems with this change when tested inworld

KristenMynx:
I think per the protocol someone other than me needs to merge.

Issue #789, opened 2022-05-17
Login or wearing collar and cuffs excessively chatty.


KristenMynx:
What version of OpenCollar are you using?  8.2.1

What behavior did you expect?
[12:48] Kristen's @ - Silky Ribbons Open Collar: Collar is preparing to startup, please be patient.
[12:48] Collar: Addon connected successfully: Lower Left Arm Cuff
[12:48] Collar: Addon connected successfully: Leg Chains
[12:48] Collar: Addon connected successfully: Lower Left Leg Cuff
[12:48] Collar: Startup in progress... be patient
[12:48] Collar: RLV ready!
[12:48] Collar: Addon connected successfully: Arm Chains
[12:48] Collar: Collar ready. Startup complete

What behavior did you see instead?
[12:48] Kristen's @ - Silky Ribbons Cuff Right Wrist: oc_cuff_pose ready (40294b)
[12:48] Kristen's @ - Silky Ribbons Cuff Right Ankle: oc_cuff_pose ready (40294b)
[12:48] Kristen's @ - Silky Ribbons Open Collar: Collar is preparing to startup, please be patient.
[12:48] Kristen's @ - Silky Ribbons Cuff Left Wrist: Lower Left Arm Cuff ready (4614b)
[12:48] Collar: Addon connected successfully: Lower Left Arm Cuff
[12:48] Kristen's @ - Silky Ribbons Cuff Left Ankle: Lower Left Leg Cuff ready (4388b)
[12:48] Kristen's @ - Silky Ribbons Cuff Right Ankle: Leg Chains ready (3554b)
[12:48] Collar: Addon connected successfully: Leg Chains
[12:48] Collar: Addon connected successfully: Lower Left Leg Cuff
[12:48] Collar: Startup in progress... be patient
[12:48] Collar: RLV ready!
[12:48] Kristen's @ - Silky Ribbons Cuff Right Wrist: Arm Chains ready (3894b)
[12:48] Collar: Addon connected successfully: Arm Chains
[12:48] Collar: Collar ready. Startup complete
[12:49] Collar: Addon: Lower Left Arm Cuff has been removed because it has not been seen for 5 minutes or more.
[12:49] Collar: Addon: Lower Left Leg Cuff has been removed because it has not been seen for 5 minutes or more.
[12:49] Collar: Addon: Leg Chains has been removed because it has not been seen for 5 minutes or more.
[...]

PR #792, opened 2022-05-24
Kristen mynx addon less chatty patch

KristenMynx:
Changes to implement Issue #789.   Login or wearing collar and cuffs excessively chatty.  I'm the only one advocating for this, but I think its needed.


tayaphidoux:
while this is good and your code looks like it will work maybe make it a opt in debug toggle under the addons menu where you can set other special settings

tayaphidoux (approved):
simple enough it should work with no problems, as i mentioned in a comment a Setting->Addons menu opt in would be a good idea for creators.

Pingout (commented):
Works great, thank you

Issue #773, opened 2022-04-27
Auto teleporting not working


ClaireBare:
What version of OpenCollar are you using?  OpenCollar 8.2.1000

What behavior did you expect? That when I tp my pet he comes whether at keys or not

What behavior did you see instead?  he has to be at the keys.

What steps does someone need to take to reproduce the problem?
No one else seems to have the problem 
I have checked - his RLV is on
the area where I teleporting from has scripts on
I have checked under RLV and Exceptions like told in world group and the auto tp is ticked.  
If I untick it close the window down and then open and retick it works once but not again

tayaphidoux:
i have experienced this and have not figured out exactly what causes it beyond relay devices, the other times it seems to be lost due to lag often times when teleporting or a slow login.

ClaireBare:
> i have experienced this and have not figured out exactly what causes it beyond relay devices, the other times it seems to be lost due to lag often times when teleporting or a slow login.

I am not sure at all what causes it, like I said, if I go into Exceptions, untick it, close it down open it up again and re-tick it, it will work one time and then I have to do it again, fortunately he is at keys most of the times but sometimes not

Medea-Destiny:
"whether at keys or not" -- @ClaireBare could you check to see if this is happening _specifically_ when he is AFK?

ClaireBare:
> "whether at keys or not" -- @ClaireBare could you check to see if this is happening _specifically_ when he is AFK?

It used to happen only when he was AFK but I know that is from his viewer being minimized, it happens all the time now, he has to physically click the tp request, so if he happens to step away from the computer for 5 mins I cannot tp him (if say the sim resets and I want to tp him away, I cant)

tayaphidoux:
> "whether at keys or not" -- @ClaireBare could you check to see if this is happening _specifically_ when he is AFK?

this happens to me often even when i am not afk and it only really started becoming common after sl's move to cloud, there is also issues going on with people being shown as offline in users profiles when they are online and vise versa, its an sl bug and could be related if afk affected it in the past.

ClaireBare:
this happens to me often even when i am not afk and it only really started becoming common after sl's move to cloud, there is also issues going on with people being shown as offline in users profiles when they are online and vise versa, its an sl bug and could be related if afk affected it in the past.

Glad to know I am not the only one, have you found a solution or work around Taya?

tayaphidoux:
[...]

Issue #753, opened 2022-03-15
Owner and .setting restrictions removed by rlv objects


PonySorrowsong:
What version of OpenCollar are you using? 8.2.1

What behavior did you expect?
I expected to have my owner restrictions restored when I used muffin milkers. But it imposed many of the same restrictions until the daily quota is made, then removes them. But since they were the same as my owner's restrictions, it removed his as well. 

What behavior did you see instead?
This is what the message is when Quota is achieved:

[04:48] MM - Milker Stall v.1.46 whispers: Quota Reached. Unlocking....
[04:48] PLM Open Collar 8.2.1: @unsit=y
[04:48] PLM Open Collar 8.2.1: @clear=tplocal
[04:48] PLM Open Collar 8.2.1: @clear=tplm
[04:48] PLM Open Collar 8.2.1: @clear=tploc
[04:48] PLM Open Collar 8.2.1: @clear=tplure_sec
[04:48] PLM Open Collar 8.2.1: @clear=showworldmap
[04:48] PLM Open Collar 8.2.1: @clear=showminimap

What steps does someone need to take to reproduce the problem?
There are several solutions I can see. 1) Store a list of of who imposed a restriction. if the restriction was created by an owner or trusted person, then a scripted object not owned by an owner should not over-wright the owners permission or be able to remove it when the object removes it. there should be an orderings of restriction, where the owners or other persons restrictions would have a higher priority when a lower priority object exerts temporary restrictions that are removed. 
2) if that is too complicated, put a single line in an oc script someplace that basically says, " if (un-sit | stand) Load owner commands stored in some list, or from the .settings notecard if they are active and unchanged by an owner who changed .settings using them menu.

mistressohm:
Keep in mind that the collar is not what is setting the restrictions here.
It is only reporting on what its relay is being told to do.

The MM Milking Stall is imposing all restrictions, if (!) the owner of that
stall is trusted by the collar's relay.  To complicate matters, if you wear
the full version of the MuffinOS HUD, it has its own relay in it, pre-set
to work ONLY with milking stalls.

Neither the stalls nor the HUD query for any exceptions stored on any
collar, OC or otherwise.   Since Muffin does not work with OC to keep
things 'compatible' (and is not expected to), from an OC standpoint the
only solution here is to not trust the landowner that put the stalls in
place.

PonySorrowsong:
[...]

Issue #725, opened 2022-01-24
Auto Relay not working after object capture & release


FrankkParisi:
What version of OpenCollar are you using?
8.2.1

What behavior did you expect?
Relay is set to auto, helpless, trust owner (on menus and in .settings)
I am the owner of her collar
I would expect her to be auto tp'd to me without asking at all times unless(?) she is currently locked on an rlv enabled object.
If she has been locked and./or released then auto tp should resume

What behavior did you see instead?
Sending a TP immediately TPs her to me in normal circumstances 
BUT..
If she has been rlv captured by an object (e.g. chair, cage etc) and then manually released (i.e release, stop etc .. ) Relay behaves as if Ask is selected... So i can no longer auto tp her unless she logs or removes and replaces collar, have tried clear RLV - no good, have also tried load (.settings), no good.
I have tried this on several objects and all exhibiting the same problem

What steps does someone need to take to reproduce the problem?
Follow the steps above

Issue #722, opened 2022-01-18
Owner Exception Receive IM not working on v 8.2.1    


jess4441:
What version of OpenCollar are you using?

What behavior did you expect?

What behavior did you see instead?

What steps does someone need to take to reproduce the problem?

jess4441:
Expected to be able to receive IM from Owner when IMs were restricted but instead received this .. .
[14:25] jess (jesssub Resident): test
[14:25] Nora (DawnStarbright Resident): *** IM blocked by your viewer  

all boxes on the Owner exception menu are checked as per gyazo https://gyazo.com/63a43e7ec98ac997bc5cb8af77601444

PR #794, opened 2022-05-25
Fix: Relay clearing all restrictions and exceptions of the collar upon receiving !release or @clear

KristenMynx:
Many comments regarding this (by me) are in #753.    This fix could potentially also fix: #773, #725, #722.

The relay is changed to send all RLV commands through RLV_CMD link messages so the rlvsys can handle them.  oc_rlvsys then arbitrates between collar restrictions and exceptions and relay restrictions and exceptions.   Previously the relay would send its rlv commands directly to the viewer.

Also removed DO_RLV_REFRESH and changed the way the release functions.


tayaphidoux:
@KristenMynx it does fix the issue but does not clear the relay set restrictions the solution for this is instead of @clear when its called keep a list in the relay script of the settings requested, and parse through it undoing each one rather than just passing the @clear.

tayaphidoux:
@KristenMynx further study suggests that instead of doing away with DO_RLV_REFRESH hunt down why RLV_REFRESH is not triggering a the default settings to re initialize!

tayaphidoux:
can you explain what removing else if(iNum == TIMEOUT_FIRED)
        {

            if(sStr == "recheck_lock"){
                if(!g_iCollarLocked){
                    llOwnerSay("@detach=y");
                }
            } has to do with it

tayaphidoux:
@KristenMynx found a solution

`
RLV(string cmd,key kID)
{
    if(kID == NULL_KEY || kID == "" ){
        if( Source == NULL_KEY || Source == ""){
            kID == "a94c4e13-ff4f-4686-b14f-877c6424d9d8";
        } else {
            kID == Source;
        }
    }
    // send the rlv command to oc_rlvsys
    llMessageLinked(LINK_THIS, RLV_CMD,
        llGetSubString(cmd,1,999),  // knock off the "@"
        kID);
    //    "a94c4e13-ff4f-4686-b14f-877c6424d9d8");
    // we use a UUID as the source because oc_rlvsys determines
    // if this is an external object or an internal script which sent the RLV_CMD.
    // since we're sending this internally on behalf of an external object we use a UUID
    // eventually we should pass the actual object, so oc_relay can then handle multiple relay objects
    // because oc_rlvsys seems to be able to handle/abritrate/manage multiple sources.
}
`
and every where RLV() is called use Source or the available id in places where source is not yet set.

tayaphidoux (changes requested):
needs just a small change and it will work as intended i proposed the change in the comments we can change it here or submit it and i can make a pull request.

KristenMynx:
@tayaphidoux   Regarding 'why RLV_REFRESH is not triggering a the default settings to re initialize!'   RLV_REFRESH is not supposed to do that.  The purpose of RLV_REFRESH is to inform all plugins/addons that they should reinstate their RLV state (through RLV_CMD) because something happened (login, rez, reset).   It is not a command to rlvsys to refresh the state of things.   rlvsys does that automatically upon startup.   RLV_REFRESH should not cause the default settings to reinitialize.

It is the responsibility of each plugin/addon to watch for *SETTING* link messages and set their internal variables.   Settings are pulled when rlvsys startes up, so all plugins should get their settings as a side effect of rlvsys pulling.    No plugin should need to as for settings on their own.  

The same applies to RLV_VERSION and RLVA_VERSION.   rlvsys informs every plugin/addon through those link messages of the rlv version(s).  It does this on every startup (wear/rez/reset/inventory change).   There should never be a need for a plugin to ask for that information, simply notice the messages and store the result as it needs.

KristenMynx:
All that said, in my view, the changes as offered need testing.

corvannansen:
My girl and I encountered this error with the "CLEAR ALL" command in the RLV menu. The fix worked, and exceptions were not cleared when other restrictions were.

tayaphidoux (approved):
still would like to see changes added but if they aren't that can be solved later, we have had a regular user prove that the changes do solve the problems expressed.

KristenMynx:
@tayaphidoux  I totally agree that your idea should work.   I'm simply taking a conservative approach of not releasing a new feature (handle multiple devices) at the same time of a philosophical change in the way that rlv calls and rlv state are being handled by the relay.

tayaphidoux:
@KristenMynx 
1) without the change your code falls short of working.
2) its already got the filter to prevent more than one device operating at the.
3) this is in no way trying add the new feature but fix a bug.

KristenMynx:
@tayaphidoux Can you clarify the change you'd like then.   I'm totally missing something then.

KristenMynx:
@tayaphidoux Ok i see it is not a feature add.   I do fail to see how it falls short of working.  

Sending Source through instead of a constant UUID seems safer.   oc_rlvsys will see a the relay as a single entity, regardless of what devices appear and disappar throughout the course of the relays use over time.    That seemed safer to me, until we can code the relay be multi-device.

tayaphidoux:
@KristenMynx  i have said this four times now.
when an object engages with relay it works correctly, when you end engagement in the case of the muffin milkers  which i tested on it does not release the restrictions, none at all what so ever, which prevents it from clearing collar restrictions but breaks the relay's functionality.

KristenMynx:
@tayaphidoux Thank you!

KristenMynx:
@tayaphidoux I have heard of the concept, but have no experience.

tayaphidoux:
@KristenMynx  they apply the following restrictions through the relay

### RLV RESTRICTIONS #####################
### TE - Paws Collar Safra's Pony - chest - test (Neck) #####################
-------- unsit
-------- tploc
-------- tplure_sec
-------- showworldmap
-------- showminimap
##########################################

SilkieSabra:
I see there is an approving review but there are two coders in this conversation and both of you have contributed code to the same pull as far as i can tell.  Can we get an inworld script you both agree on that can be tested?  I think this could be the last piece we need to push an update that will support the AO and Cuffs.

KristenMynx:
@SilkieSabra I did a bunch of testing with the script is pulled from the net result of my pull request on the device that @tayaphidoux  tested on.    We worked together, but I was unable to make it fail.   One possibility was that she's using ask mode.   So I came back today to test with ask mode.    Things worked properly in a sense, but I found a new bug in ask mode.   In ask mode, only the first restriction is accepted, all others are ignored.   I found a fix for this bug and will be doing another change to my pull request to fix that.    I discovered yet another bug... The relay tries to remember a source that was granted access (the last 5 since rez/wear) and not bother to prompt again.   This doesn't appear to work.  If its an easy to fix and low risk change I'll do that too.

KristenMynx:
@tayaphidoux @SilkieSabra   I've sent you oc_rlvsys PR 794 and oc_relay PR 794 in world.
Testing should be done for the original bug, plus Phidoux's issue: when a device releases you, not all those restrictions are removed.    Also test if Ask mode adds (during capture0 and removes (during release) all the restrictions that the device requested.

tayaphidoux:
@KristenMynx  i will have to try it when i next get on that would be the optimal time, one other reason i may not work  for me is i am using kokua which is RLV not RLVA but that don't explain why getting the key from the object works where your key does not, however i will give your scripts a try

tayaphidoux:
@KristenMynx had an issue where i had to re log because the relay kept the last item to ask for relay permission when i got off it enough time didn't pass for the new item to be able to request permission, i am testing weather it releases on its own today

JennGoodman (approved):
[Kitty(sweetdangerkitty) ]The code is good. Just need to resolve the merge conflict.

Pingout (approved):
Relay and RLVsys both working well

Issue #772, opened 2022-04-14
Custom exceptions (done badly) can cause odd IM distance modifiers


DeaEnyo:
What version of OpenCollar are you using? 8.2.1

What behavior did you expect? Adding a custom exception and checking any of the IM boxes adds that exception.

What behavior did you see instead? Due to sloppy copying of an avatar's UUID from their profile, I missed the last character. Upon activating the IM exceptions I was subject to a modifier limiting the minimum IM distance. (from 25m up to 8845178m across several malformed UUIDs)

What steps does someone need to take to reproduce the problem? Be bad at copying an UUID into a custom exception, apply IM exceptions, watch the Modifiers tab of RLVa.

PLEASE do a sanity/idiot/length check on each key pasted into the avatar UUID entry dialog.  There is no indication that anything is wrong and it took me several hours of dopey fiddling to track down the issue.

mistressohm:
I would actually recommend that a user lookup be performed on any UUID
submitted.   One doesn't have to save the data returned, just know that the
lookup was successful.

Not only does it rule out 'malformed' data (since the lookup would fail)
but also ensure that owners / trusted / blocked persons actually exist
within SL.  This means that you're not using script memory to store data
that is invalid or obsolete.

There's also the issue of 'ghost' owners - if you've a locked collar, do
not 'own' yourself, and your Owner deletes their SL account, what happens?

DeaEnyo:
That would be an excellent solution to many of the rare but serious problems that could occur. 
I know at least one person who had lost their Owner to a ban and was forced to cheat out of their collar.
(I should explain HOW I messed up a simple action so badly: Instead of 'copying URL' from the resident's profile and deleting the "secondlife:///app/agent/" and "/about" parts, I simply dragged mouse cursor over the Key field to select it all, then copied. It turns out that some characters take up more space than others, so while it works on some keys, such as mine, it will miss the last one on others. A Firestorm/viewer UI issue, essentially. )

mistressohm:
Yah I prefer using the Key field myself, but I'm also a big fan of
triple-clicking (which should highlight the whole thing) or CTRL-A (which
"selects all"). Then either right click and copy or CTRL-C.

Sanity checks are never 'wasted code'.

KristenMynx:
[...]

PR #796, opened 2022-05-26
Verify Custom Exception UUID is a proper UUID

KristenMynx:
Does not check if the UUID is a valid Avatar nor a valid Group.   Reference #772


tayaphidoux (approved):
i have had to do a check like this once in the past it looks like would work and so far does not break the collar.

Issue #767, opened 2022-04-09
Collar Auth behaviour not consistent when accessed by AO's "Collar Menu" button


MuttMode:
What version of OpenCollar are you using? 8.2.1

What behavior did you expect? 
     I am an owner on my own collar. I click, Access, + Owner and expect to be asked who to add

What behavior did you see instead? 
     I click + Owner and get "Access Denied to adding a person".  I can 'force' it to recognize my authority (!) by going into RLV, Relay and altering a setting. Then back to Access and I can add/ change settings as expected.

What steps does someone need to take to reproduce the problem? 
     Wear a 8.2.1 collar, add self as Owner. Log off, log on, wait for collar to boot up, try to add an actual Owner *WITHOUT* going into other collar options first.

mistressohm:
What options specifically do you have to change?  Does any setting in RLV
Relay do this, or only specific ones?
Do you just toggle the setting(s) On and then Off again?  or do you have to
turn something on or off, leave it there, and go back to add an Owner?

You might need to dump your Settings into local chat, and copy them into
the Github ticket.  You can sanitize by replacing your own UUID with [my
UUID], and the UUID's of trusted/owners with either [Trusted UUID] or
[Owner UUID].

mistressohm:
Also, it would be good to know if you're under any RLV / relay restrictions
at the time you attempt the above.  You should be able to get them from the
Firestorm RLVa menu, under "Restrictions".  There's also a chat command
to dump them into local chat, but I can't recall it offhand.

MuttMode:
There were no restrictions apart from the collar's own lock though I don't think that is relevant to this issue (I am often wrong!)
Tested with a fresh, unmolested 8.2.1 from the OC HQ:
Wore collar, added self as Owner, added another Owner.
Removed collar. Wore collar, "You are owned by: Yourself, OtherOwner" message came up as expected.
Click AO, clicked "Collar Menu", then "Access" then "+ Owner".
"OpenCollar: Access Denied to adding a person"

After a few further tests it seems that my habit of accessing my collar via the AO is the issue. DIRECTLY clicking the collar gives me the correct auth privileges. 

Summary/correction to original issue:
Accessing collar via AO does NOT recognize the wearer's auth level correctly.

mistressohm:
Ok, thanks. That narrows things down a bit.    The AO may need some
tweaking to match up with changes to the latest version.

Are you able to edit the Title of this Github ticket to match that?

[...]

PR #797, opened 2022-05-26
'Collar Menu' to allow normal wearer functions

KristenMynx:
Changed the message number from CMD_WEARER (503) to CMD_ZERO (0).  This causes the collar to validate the access level by the toucher uuid, and restores proper functionality to things like adding owners, etc when the wearer is also an owner.   Reference #767


tayaphidoux (commented):
this issue should be resolved in the changes i have made to the ao.

KristenMynx:
@tayaphidoux The collar menu option is handled differently in your version.  This PR is therefore moot.   Closing.

PR #798, opened 2022-05-26
Make RLV_VERSION call when not using RLVa.

tayaphidoux:
this is not a critical update but may be use full in some projects and cut down on chatter spam by not requiring the all in the event you want to reconfirm rlv. and also puts RLV_VERSION to work to confirm legacy rlv if RLVA_VERSION is false.

### DIFF
--- a/src/collar/oc_rlvsys.lsl
+++ b/src/collar/oc_rlvsys.lsl
@@ -69,6 +69,7 @@ integer DO_RLV_REFRESH = 26001;//RLV plugins should reinstate their restrictions
 integer RLV_CLEAR = 6002;//RLV plugins should clear their restriction lists upon receiving this message.
 integer RLV_VERSION = 6003; //RLV Plugins can recieve the used RLV viewer version upon receiving this message..
 integer RLVA_VERSION = 6004; //RLV Plugins can recieve the used RLVa viewer version upon receiving this message..
+integer RLV_VERSION_REQUEST = 6005; //RLV Plugins and addons can exclusivly request which version was detected.
 integer RLV_CMD_OVERRIDE=6010; //RLV Plugins can send one-shot (force) commands with a list of restrictions to temporarily lift if required to ensure that the one-shot commands can be executed
 
 integer RLV_OFF = 6100;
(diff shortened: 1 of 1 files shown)


KristenMynx:
I'm not sure I see a need for this.  Every script will receive RLV_VERSION and RLVA_VERSION as a link message when the collar starts up (attached, rezzed).   Any script that needs the version should simply be listening for it, and store it for future use.

tayaphidoux:
@KristenMynx also at my research the number i used for the all is currently used by no other official script, so it don't conflict with that, secondly it just makes this script accept the call to receive version individually when you only want the version.

KristenMynx:
@tayaphidoux I still see no point to this.  Each script should receive a link message with RLVA_VERSION and RLV_VERSION upon collar wear/rez/reset/inventory change (rlvsys does this)  There should be no reason for an addon to ask for the version.   It should simply store the version as needed in a variable and consult that variable.

tayaphidoux:
@KristenMynx without this it only receives the one RLVA_VERSION, as for the call  to trigger it there could be situations you are between those situations and you just need the one parameter?

KristenMynx:
@tayaphidoux It is true that when rlvsys receives SETTINGS, "ALL", it only sends RLVA_VERSION.  But in my view is not what a plugin script should be relying on.   Perhaps an appropriate fix is to send RLV_VERSION at that time as well, but it is really too early in the process for the version to be accurate.

state_entry calls setRlvState().   on_rez does a reset, so that also calls it.  So when the collar is worn, setRlvState() is called.   Also a settings, "ALL" message is sent.   

When a setting "rlvsys" "on" is received, setRlvState() is again called since now the script knows rlv is turned on.

setRlvState() starts a process with a timer, setting g_iWaitRelay=1.  setRlvState starts a version check, in concert with the timer().  When the timer times out with g_iWaitRelay set, both RLV_VERSION and RLVA_VERSION messages are sent.

This means that any wearing/rezing will cause both version messages to be sent.
I do see that inventory change (LM_SETTINGS_REQUEST,"ALL", which is done in the core script) will only send RLVA_VERSION.

Again, perhaps the proper fix is to change rlvsys to send both versions on the LM_SETTINGS_REQUEST, "ALL" message, which would include wear/rez/inventory change.

During your testing where you notice that only RLVA_VERSION is sent, were you wearing/rezing, or just changing inventory/compiling a script.   If the latter, I can understand why you aren't seeing the RLV_VERSION message you expect.

If that is the case, the change I suggested would be the proper fix.

tayaphidoux:
@KristenMynx  you have just confirmed that you aren't actually reading the changes i made 

you are taking issue with this

```
	} else if(iNum == RLV_VERSION_REQUEST ){
			if (g_iRlvaVersion) {
               	llMessageLinked(LINK_SET, RLVA_VERSION, (string) g_iRlvaVersion, NULL_KEY);
			}
            else {
				llMessageLinked(LINK_SET,RLV_VERSION, (string) g_iRlvVersion,NULL_KEY);
			}
```

and missing the change from this
```
                if (g_iRlvaVersion) llMessageLinked(LINK_SET, RLVA_VERSION, (string) g_iRlvaVersion, NULL_KEY);
            }
```
to this
```
                 if (g_iRlvaVersion) {
                	llMessageLinked(LINK_SET, RLVA_VERSION, (string) g_iRlvaVersion, NULL_KEY);
                }
                else {
					llMessageLinked(LINK_SET,RLV_VERSION, (string) g_iRlvVersion,NULL_KEY);
				}
			}
```

which satisfies your comment here 
" Again, perhaps the proper fix is to change rlvsys to send both versions on the LM_SETTINGS_REQUEST, "ALL" message, which would include wear/rez/inventory change."

the part you have issue with i am doing for the sake of addons, and sanity checking in situations where rezing/wearing compiling and changing inventory does not always happen at a timely interval and an addon script especially one not worn could potentially miss the call, like for example a dresser addon for rp situations and reasons of limiting outfit changing to a single location.

"During your testing where you notice that only RLVA_VERSION is sent, were you wearing/rezing, or just changing inventory/compiling a script. If the latter, I can understand why you aren't seeing the RLV_VERSION message you expect."

SilkieSabra:
I think we do need this.  Can you compile it inworld and pass copies to Ping and Omnia for testing? What should be a proving inworld test?

tayaphidoux:
@SilkieSabra  will do when i get a chance to work on things and my view is not bouncing about so that i can build

Pingout (approved):
I had no problems with this inworld.
@SilkieSabra this can be merged

PR #777, opened 2022-05-05
Update oc_unwelder.lsl

Pingout:
Ref Issue #776

PR #804, opened 2022-06-14
make welder more stable when connecting to collars

tayaphidoux:
this change is based from conversation in open collar R&D about welder not staying connected or randomly dropping out, this fix should hopefully help keep connection, it works for my add-ons and the AO changes i have been working on and i have seen much less chatter about add-ons not being seen.

### DIFF
--- a/src/spares/oc_unwelder.lsl
+++ b/src/spares/oc_unwelder.lsl
@@ -150,7 +150,7 @@ default
             
     //        llSay(0, "Unwelder now in use");
             llSetText("In use by\n" + llKey2Name (llDetectedKey(0)), <1,0,0>,1);
-            llSetText("In use..", <1,0,0>,1);
+            //llSetText("In use..", <1,0,0>,1);
             g_iLMLastSent = llGetUnixTime();
             g_kUser=llDetectedKey(0);
             API_CHANNEL = ((integer)("0x" + llGetSubString((string)g_kUser, 0, 8))) + 0xf6eb - 0xd2;
--- a/src/spares/oc_unwelder.lsl
+++ b/src/spares/oc_unwelder.lsl
@@ -116,11 +116,28 @@ integer g_iLMLastSent;
 key g_kUser = NULL_KEY;
 integer g_iWelded=FALSE;
 integer g_iAddonLimitation = TRUE;
+integer softCounter = 0;
+
+softreset(){
+   if(softCounter < 3){
+      softcounter ++;
+      // assume unintentional disconnect and try to salvage.
+      g_kCollar = NULL_KEY;
+      API_CHANNEL = ((integer)("0x" + llGetSubString((string)g_kUser, 0, 8))) + 0xf6eb - 0xd2;
+      Link("online", 0, "", g_kUser);
+   }
+   else {
+      // if we have tried enough times do a hard reset because the item may no longer be present!
+      llResetScript();
+   }
+}
+
 default
 {
     state_entry(){
         llSetText("",ZERO_VECTOR,0);
         llSay(0, "Unwelder is now ready");
+        llSetTimerEvent(10);
     }
     touch_start(integer t){
         if(g_kUser != NULL_KEY){
@@ -140,7 +157,6 @@ default
             llListen(API_CHANNEL, "", "", "");
             Link("online", 0, "", g_kUser); // This is the signal to initiate communication between the addon and the collar
             llResetTime();
-            llSetTimerEvent(60);
             
         }
     }
@@ -149,7 +165,8 @@ default
     {
         if(llGetTime() >= 30.0 && g_kCollar==NULL_KEY && g_kUser!=NULL_KEY){
             llSay(0, "Connection timed out to collar.");
-            llResetScript();
+            softreset();
+            //llResetScript();
         }
         if (llGetUnixTime() >= (g_iLMLastSent + 30) && g_kCollar != NULL_KEY)
         {
@@ -160,7 +177,8 @@ default
         if (llGetUnixTime() > (g_iLMLastRecv + (5 * 60)) && g_kCollar != NULL_KEY)
         {
             g_kCollar = NULL_KEY;
-            llResetScript(); // perform our action on disconnect
+            softreset();
+            //llResetScript(); // perform our action on disconnect
         }
         
         if (g_kCollar == NULL_KEY) Link("online", 0, "", g_kUser);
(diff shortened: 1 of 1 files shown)


tayaphidoux:
discord "
[2022/06/13 15:28] ròan (silkie.sabra): we are still having problems with the unwelder not always connecting to AddOns too, sometimes there just is no popup to start the unweld process... first look at it decided it was people trying to unweld using peanut collars but that's not always the case and it appears to be random
[2022/06/13 15:30] Phidoux (taya.maruti): i would have to look and see what the initializer for the welder is, the addons drop off quite often, and might be a solution to droping their timer to 10s it uses llGetUnixTime() to validate the difference for larger, however we have it set to 60s and some of the checks and pings are less than that
[2022/06/13 15:31] Phidoux (taya.maruti): the addons also try to connect using the object key which changes if you relog
[2022/06/13 15:31] Phidoux (taya.maruti): in the case of connecting to the collar its the collars key
[2022/06/13 15:32] Phidoux (taya.maruti): secondly while llRegion say is possible once the addons are linked to the collar they only listen to the collar
[2022/06/13 15:32] Phidoux (taya.maruti): speak only to i mean
[2022/06/13 15:33] Phidoux (taya.maruti): so it could be a matter of one of your requirements not being able to be met except in the right condition
[2022/06/13 15:33] Phidoux (taya.maruti): i am having the same situation with the ao menu right now
[2022/06/13 15:33] Phidoux (taya.maruti): if the listen is bussy it drops the menu
[2022/06/13 15:34] ròan (silkie.sabra): unfortunately it's beyond me so at least in the discord your information will be saved
[2022/06/13 15:37] Andur Faulds: does the collar buffer the incoming speech? it might come down to how many voices are yelling at once, in what order, and the collar can only listen and process so many. Increase the buffer size or sequence the order of possible calls to a synchronized rhythm
[2022/06/13 15:39] Phidoux (taya.maruti): that is a nother possible solution make the checks longer before it kicks the addons to cut down on chatter, that would have to be changed both in the collar side and addon template
[2022/06/13 15:39] Phidoux (taya.maruti): and you would risk having issues with addons taking to long to refresh if you modified them or detached them
[2022/06/13 15:41] Phidoux (taya.maruti): i have been trying to get the addons to announce when they disappear but lsl is not very good at that one even though it has functions which kinda work for that.
[2022/06/13 15:42] Phidoux (taya.maruti): state_exit() and attached(key id){ id != llGetOwner()} are the two situations i know of and change link might work to
[2022/06/13 15:42] Phidoux (taya.maruti): however change link would shout any time you unsit from an object
[2022/06/13 15:43] ròan (silkie.sabra): the AddOns are way too verbose in any case
[2022/06/13 15:44] ròan (silkie.sabra): most people are just confused by announcements that something has been dropped because it hasn't been seen etc or trying and failing to connect etc
[2022/06/13 15:44] ròan (silkie.sabra): it's like debug messages were left in there
[2022/06/13 15:44] Phidoux (taya.maruti): yes i noticed the changes for that reason
[2022/06/13 15:45] Phidoux (taya.maruti): i have been trying to find a way to solve it by fixing the connection timers and exit calls because not only do you get that message but when it happens all of your addons are nonfunctional till they reload
[2022/06/13 15:45] Phidoux (taya.maruti): its supposed to only kick the offending item but it kicks them all
[2022/06/13 15:51] Andur Faulds: *you all offend me! get out!* Lol, i can understand that...
[2022/06/13 15:53] ròan (silkie.sabra): in addition, we need a better way to assign issues like this and make sure they get tested and reviewed in some orderly way, we're working on that too
[2022/06/13 15:55] Phidoux (taya.maruti): an example of what i am talking about  if(llGetTime() >= 30.0 && g_kCollar==NULL_KEY && g_kUser!=NULL_KEY){
            llSay(0, "Connection timed out to collar.");
            llResetScript();
        }
[2022/06/13 15:55] Phidoux (taya.maruti): by the time that is pinged it will aways be > 30
[2022/06/13 15:55] Phidoux (taya.maruti): with the llSetEventTimer(60)
[2022/06/13 15:56] Phidoux (taya.maruti): makes the time check portion useless
[2022/06/13 15:56] ròan (silkie.sabra): so the thing to do would be first file an issue, describe the problem, then offer a fix in a pr, then someone has to make sure it gets tested and reviewed so it can be merged.  That "someone" i think is what they would call a point guard in basketball
[2022/06/13 15:57] ròan (silkie.sabra): and that's what we're missing
[2022/06/13 15:57] ròan (silkie.sabra): in terms of getting things moved forward in a timely way
[2022/06/13 15:59] Phidoux (taya.maruti): in the case of the welder move the timer to state_entry if there is no calls ending it
[2022/06/13 16:00] Phidoux (taya.maruti): keep the llResetTime() function  and llGetTime() portion won't be entirely usless
[2022/06/13 16:01] Phidoux (taya.maruti): approved should update g_iLMLastRecv
[2022/06/13 16:01] Phidoux (taya.maruti): that will help with a little with keeping the ping alive
[2022/06/13 16:05] Phidoux (taya.maruti): from the looks of it and maybe a softreset instead of llResetScript(); with the welder retrying to connect if it times out, can also use a counter if it fails 3 times to reconnect disconnect
[2022/06/13 16:06] Phidoux (taya.maruti): in the event that it gets thrown off by a bad addon
[2022/06/13 16:06] Phidoux (taya.maruti): by safe reset i mean you reinitialize the connection without changing the settings
[2022/06/13 16:07] Phidoux (taya.maruti): softreset *
[2022/06/13 16:07] Phidoux (taya.maruti): i have used this method and it hs reduced the reset calls
[2022/06/13 19:21] Phidoux (taya.maruti): i proposed the changes i suggested for the welder you can review them and test them at your discretion.
[2022/06/13 19:23] ròan (silkie.sabra): thank you PHidoux"

Pingout:
It's a little confusing between this PR and PR#805. I originally reviewed *this* script but put my comments in the PR #805 
PR #805 is the Unwelder HUD, where as this is the Unwelder Device Script
That being said, we do have a slight problem in this script which was previously addressed in PR #777 
In a crowded location, like the OC Headquarters, it's not enough to know the Unwelder is "in use", we need to know who is currently using it. Line 152-153 section uses a llSetText command to indicate who is using the device. The script appears to use the llSetText twice in a row, and I suspect the second command overrides the first.
   Line 152           llSetText("In use by\n" + llKey2Name (llDetectedKey(0)), <1,0,0>,1);
   Line 153           llSetText("In use..", <1,0,0>,1);
We should comment out the second line.

tayaphidoux:
@Pingout  that should over that now the first llSet text will not be over written by the second.

Pingout:
This is awesome Taya, thank you.
@SilkieSabra you can go ahead and merge this PR

SilkieSabra:
@PIngout where's the approval check in the review?

Pingout (approved):

PR #807, opened 2022-06-15
a better merger of addon and ao

tayaphidoux:
i have done some extensive work merging the add-on code with the ao and reducing the chance of a stack heap collision as well as speeding up menus,
1) i have split the original script into two the menu system and the animator
2) i have merged the addon code replacing the old code that as intended for connection.
3) i have made it possible for some of the commands to be called from comandline /1prefixAO command
3.A) commands are as follows lock,unlock,load,on,off and menu though the last one is not needed.
4) pull request for the new script https://github.com/OpenCollarTeam/OpenCollar/pull/808

known bug:
the menu occasionally drops out while in use if the collar sends a message as you click the buttons.
standing up form an item may prevent your stand/walk animations from updating, this bug happens with the other version as well.

### DIFF
--- a/src/ao/oc_ao.lsl
+++ b/src/ao/oc_ao.lsl
@@ -257,7 +257,7 @@ ToggleSitAnywhere() {
         if (g_iSitAnywhereOn) {
             llMessageLinked(LINK_THIS,AO_SETOVERRIDE,"switchstand",llGetOwner());
         } else {
-            llMessageLinked(LINK_THIS,AO_SETOVERRIDE,"set_Standing="+g_sSitAnywhereAnim,llGetOwner());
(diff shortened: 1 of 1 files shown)


Pingout:
I'm getting an error on testing, when i click the sit icon i see an error: [20:02] Could not find animation 'oc'

tayaphidoux:
@pingout i will look into it

Pingout:
> 

Hi Taya, yes all of the default animations in the current AO use an underscore character in the name

tayaphidoux:
@Pingout  then you may notice i changed the scripts to fix that very issue the problem was i wasuing _ to separate some of the calls between the two so i changed it to : which should be far less common

Pingout:
Extensively tested by Ping and Omnia, and Taya helped us fix a little bug, thanks!
@SilkieSabra this is ready to merge

Pingout (approved):
The improvements to this AO are all working great.

PR #812, opened 2022-06-24
make ao respond to an off signal from oc_anim

tayaphidoux:
restore the animation scripts ability to enabled/disable the ao.


Pingout (approved):
this checks out very well inworld

PR #814, opened 2022-07-11
Fixing Lock Guard Compatability

tayaphidoux:
we had the order of lock guard commands wrong target comes before command
this issue was discovered while discussing in R&D with Toy(toy.wylie) and Kiha Aridian trying to solve Kiha's issue.
https://wiki.secondlife.com/wiki/LSL_Protocol/LockGuard


tayaphidoux:
nope i am wrong my bad

PR #815, opened 2022-07-11
Create oc_debugger

tayaphidoux:
For now i will place this as an addon I personally hope it becomes a core file at some point to manage output and debugging of the collar and all addons/plugins its a relatively simple output controller for creators so that you can get customized debug information, while taming the llOwnerSay and other outputs inside your builds so that they can be hidden away unless needed in a unified manner.

### DIFF
--- a/src/spares/oc_debugger
+++ b/src/spares/oc_debugger
@@ -1,4 +1,6 @@
 /*
+This application is not recomended for regular users.
+
 This is a Basic debugger it manages messages fromt he script that don't always need to be seen in three categories
 Info - Basic information like the addons getting kicked for inactivity.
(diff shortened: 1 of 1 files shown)


SilkieSabra:
You mean as an app in spares, based on discussion in OC Staff.  Can you add a few lines at the top of the script that state what it does the way you did there?

tayaphidoux:
@SilkieSabra that was described in the description on the pull but here you go a block in the comment section of the script at the very top

Pingout:
I agree this should be a spare. As Taya stated, this script is "for creators". The average user will never use this and it could become a source of confusion.

PR #816, opened 2022-07-12
Create addon_test_debugger

tayaphidoux:
a test tool and source so that creators have an example how to use the debugger script.


tayaphidoux:
close this so the new copy is not duplicate

PR #817, opened 2022-07-12
Create plugin_test_debugger

tayaphidoux:
a tool to allow creators to test the debugger from a plugin script format.


tayaphidoux:
removing to make room for the new pull

PR #820, opened 2022-07-14
Kristen mynx cuff resizer back button fix  Fixes #818

### LINKED ISSUE #818: I just noticed a problem in the oc_cuff_resizer
What version of OpenCollar are you using? oc_cuff_resizer

The "Back" button does not work properly.  The script uses the UPMENU variable, similar to oc_resizer script on the collar. It could be that the UPMENU routine used in the collar is not apropriate for the cuffs.

What steps does someone need to take to reproduce the problem? Set up cuffs using the 2022 Q1 Features branch scripts (the most current version)

Attn: KristenMynx

KristenMynx:


KristenMynx:
MISTAKE closising

PR #822, opened 2022-07-16
fix a few bugs and bring more inline with style

tayaphidoux:
this update is to fix a few issues with persistence and clean up the code a little 

0) set the debugger to default active.
1) adding visual styles like default collar check boxes with a fall back check box style,
2) created a persistent global setting for debugger status to survive relogs and teleports, 
3) remove llReset on rez while active status has been achieve to help persistence instead we return to default,
4) cleaned up the code to be more inline with oc style guide recommends and requirements

### DIFF
--- a/src/spares/oc_debugger
+++ b/src/spares/oc_debugger
@@ -1,31 +1,13 @@
 /*
-This application is not recomended for regular users.
-
-This is a Basic debugger it manages messages from the script that don't always need to be seen in three categories
-Info - Basic information like the addons getting kicked for inactivity.
-Warning - Information about problems that won't break the collar but could cause instability.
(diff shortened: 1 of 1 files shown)


Pingout (changes requested):
So this all looks good Taya, thank you. I would like to see you change one thing.  I see you changed the App Name in line 27 (10) to remove the capitalization. 
All of the App Names in the Apps Menu are all Capitalized, so this should be listed as "Debugger"

tayaphidoux:
@Pingout  i can definetly do that the reason for the change was for chat commands but i can find a work around for it.

Pingout (approved):
Thank you very much for the quick response, Taya!
@SilkieSabra this can be merged directy to Master, as it's in the spares folder

PR #823, opened 2022-07-18
faster and cleaner code

tayaphidoux:
1) changed the code to be cleaner in line with the recommends of the style guide
2) found some ways to improve persistence of the add-ons making them harder to get rid of
3) prepped for future security changes.


tayaphidoux:
needs more testing

PR #824, opened 2022-07-18
adding the persitence and security prep

tayaphidoux:
1) cleaned up code to put it in line with the style guide.
2) added the persistence changes and prep for security changes.


tayaphidoux:
going to redo this one because somthing is funky with this pull and i don't want to risk breaking things.

PR #825, opened 2022-07-19
redoing the pr for the oc_addon_template

tayaphidoux:
the changes to fix the speed and persistence of add-ons, as well as prep for security features for the add-ons them self.
re submitting this to try and prevent the goofiness in the last attempt.

1) cleaned up code to put it in line with the style guide.
2) added the persistence changes and prep for security changes.

https://github.com/OpenCollarTeam/opencollarteam.github.io/blob/master/docs/Script-Style-Guide.md
the changes to the {} in the code are to compile with this and to make sure if statements have a clear start and stop point to help new users. at the bottom of the page you have

" The only exception is when the body of the if statement is on the same line as the if.

```
// This is tolerated, though still not recommended
if (iSomeCondition) DoAThing();
```

"

but the problem with this if statement is its hard for a new user to the code to understand where it starts and stops when surrounded by other if statements especially when its an "else if"

### DIFF
--- a/src/spares/oc_addon_template.lsl
+++ b/src/spares/oc_addon_template.lsl
@@ -4,7 +4,7 @@ This script is released public domain, unlike other OC scripts for a specific an
 -Authors Attribution-
 Aria (tiff589) - (August 2020)
 Lysea - (December 2020)
-Taya'Phidoux' (taya.maruti) - (july 2021)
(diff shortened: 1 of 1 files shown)


tayaphidoux:
there we go its now not pulling a bunch of other scripts.

tayaphidoux:
needs more testing

PR #828, opened 2022-07-25
Update oc_folders.lsl

tenshilove:
Edited descriptive text in line 299 to clarify dialog language and function of locks.


SilkieSabra (approved):
Looks good to me... next comes the approving review, can we have a coder do that because neither Anne nor I can code and we might erase a crucial bracket or comma or something

PR #829, opened 2022-07-25
Update oc_folders.lsl

SilkieSabra:
Thanks to @tenshilove fpr the original change.  I moved it from Master to features q1


tayaphidoux (approved):
simple and clear

PR #831, opened 2022-07-25
Update oc_folders.lsl

SilkieSabra:
Here it is in the correct branch


Pingout (approved):
Yup looks good, thank you @SilkieSabra

PR #832, opened 2022-07-27
bring the naming convention in line with the rest!

tayaphidoux:
the way i had named this looks odd against the rest so i am fixing that.


Pingout (approved):
This works great inworld, Taya. I tried it in three different objects and was able to get the Addon Lock menu in all three.
@SilkieSabra you can merge this into Spares when ever you want.

Issue #774, opened 2022-05-01
Conflict when Wearer is also Trusted


SilkieSabra:
In 8.2.1, you can check Wearer Trust in the Access menu.  This should allow the wearer to add Trusted and works as expected.

However, you can also press + Trust and there's an option there to select Wearer.  When someone does this, the collar treats the Wearer at the Trusted rank instead of wearer.

This means among other things that they can no longer see the Access List, or add other Trusted, and are denied Access to Runaway.

Proposed Solution:  Remove the Wearer option from the + Trusted menu.  I can see there might be a use for these restrictions but on the whole it is confusing and might be difficult to retrofit to the specific condition of Wearer is also Trusted.

Medea-Destiny:
Potentially a complicated issue. There are functions (seeing access list, runaway and others not mentioned) that wearer NEEDS to be able to access. These are currently dealt with by checking if access level is wearer. As trusted provides access to things wearer doesn't have be default, it is in theory a higher access level than wearer. Removing the option to allow wearer to be trusted is a reduction in functionality, but as we have a simplistic level system for access rather than anything granular, we would either have to retrofit a new access schema which would be a huge task (though useful) or go through all the options that allow wearer level access and do a key check to identify wearer rather than checking access level.

Can people think of situations where trusted gives wearer functions they do not have otherwise? I guess undoing restrictions that have been set by another trusted user? Anything else?

KristenMynx:
Isn't the checkmark "Wearer Trust" duplicate functionality to  + Trust, >Wearer< ?  If so removing +Trust, >Wearer< is not a reduction in functionality.  

If it is not the same, then I'll go back and see if i can decipher the difference by reading and understanding the code better.

SilkieSabra:
[...]

PR #834, opened 2022-08-03
Medea destiny core patch1

Medea-Destiny:
Fixes for issue #774. This is not a complete solution, but checks agent key match g_kWearer rather than checking auth level is CMD_WEARER throughout the script. This  should fix the issues discussed in the #774, but there are other points in a few other scripts that will have to be addressed before closing that issue.


tayaphidoux:
seems you ran into the issue for moving that i have had and it causes conflicts.

Issue #849, opened 2022-08-19
Adding Wearer to Trusted breaks WearerTrust


SilkieSabra:
8.2.2 Beta collar.  Adding Wearer to Trusted should not break WearerTrust functionality.  

When WearerTrust is checked, Wearer can add and Remove trusted.

When Wearer has WearerTrust checked and is also added to the Trusted list, wearer can no longer add and remove Trusted.

Medea-Destiny:
Fix already in my beta fixes, doing the PR now(ish).

Etaski:
Similar, likely related issue: Wearer is mentioned as "trusted" in settings NC, and can no longer use "hide" and "show", gets "Access Denied" response. Removing wearer from "trusted" entry in NC solved it.

Issue #843, opened 2022-08-17
Wearer set to Trusted Access is still denied access to Runaway in 8.2.2 beta


SilkieSabra:
We tested the 8.2.2 Beta for the fix for certain Access settings available to Wearer but not to Trusted.  In 8.2.1 and possibly earlier versions, when the Wearer is set to Trusted level access they  lose the ability to operate Access List and Runaway, since those are not available at Trusted level.

8.2.2 should fix that.  Access List is fixed.  Wearer set to Trusted can see the Access List.  However Runaway is still denied to a Wearer who is also Trusted on the collar.

Medea-Destiny:
This should be fixed now, contact me in world for testing

PR #851, opened 2022-08-20
Update oc_api.lsl

Medea-Destiny:
Fixes for #774 , #847, #843 , #849

### DIFF
--- a/src/collar/oc_api.lsl
+++ b/src/collar/oc_api.lsl
@@ -14,7 +14,10 @@ Medea (Medea Destiny)
                 -   Issue #579 Restored menuto function to interface channel for backwards compatibility
                 -   Issue #579 Added control of channel 0 listening via settings menu
                 -   Fix issue 585 give wearer accept/deny dialog to confirm runaway disable setting
-                            
+    *Aug 2022   -   Issue #843 & #774 fixes to check g_kOwner rather than CMD_WEARER. This fixes runaway
+                    menu for wearers set as trusted (fix typo in menu too). Also wearer adding to trusted/
+                    blacklist when permitted when also trusted (issue #849)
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
Testing required:

1. Compatibility of interface channel (try adding a second copy of the collar and checking the collar detaches, both with another 8.2.2 and earlier collar versions.
2. Wearer who is trusted access to runaway functionality
3. Wearer who is trusted access to adding trusted with wearertrust active
4. Other potential conflicts with wearer who is trusted.

Medea-Destiny:
Added a new fix: (prefix)prefix reset was not working as advertised in instructions. I have added the functionality and added notification to the menu user and wearer of a new prefix when changed.

Pingout:
Tested by Ping and Omnia:

Wearer who is trusted access to adding trusted with wearertrust active <-- this is not true. 
The note in the Access menu says "Wearer trust allows an owned wearer to add/remove from trusted list." <-- this is true
if the wearer is owned they can add trusted, but if they are owned and trusted they can not add a trusted

Medea-Destiny:
The problem @Pingout and Omnia observed was not actually to do with this script, but oc_core.

To clarify -- this script handles actually adding people to lists, however the menu is served from the oc_core scripts. There are fixes in the oc_core script that handle that side of things, and thus this PR needs to be tested on top of the current beta, not with the 8.2.1 version of oc_core.

Ping has confirmed that with both oc_api and oc_core updated, the trusted adding is working with both menu and chat commands.

Pingout:
@SilkieSabra this PR can be merged

SilkieSabra:
@Pingout if you say it can be merged can you please tick the Approved box?

Issue #842, opened 2022-08-17
Trap Restrictions Unwanted Behavior


Omnia2021:
What version of OpenCollar are you using? 8.2.2 Beta

What behavior did you expect? No change in ability to detach collar and no ability to detach collar

What behavior did you see instead? After releasing from trap and clearing restrictions @detach was also cleared allowing for removal of a locked collar

What steps does someone need to take to reproduce the problem? Sit on a trap be released then clear restrictions and detach locked collar

Medea-Destiny:
So... RLVsys will apply a detach if there are external sources, such as a relay. When the relay is released, the external sources become empty and the detach is removed. As the collar's lock status is not handled in rlvsys, this is a problem. I have a fixed version of rlvsys in world that re-establishes the lock if the collar was locked (which also required making rlvsys respond to an unlock as well as a lock, oops.)

HOWEVER, while I was looking through this I have noticed that we have some other issues with the relay in its current form which I want to spend a bit of time investigating, and there's a bug with the trust wearer and trust trusted buttons which don't work -- not new to this version I see, but presumably nobody noticed because the menu doesn't actually tell people what those buttons are supposed to do.

I can provide the updated RLVsys script for people to test in-world for the specific issue mentioned here which seems to be fixed now, but I'd like to do a bit of a deep dive into the relay script before it gets released into the wild.

SilkieSabra:
What is the time frame on that, should we wait 8.2.2 release for it?

Omnia2021:
You can send that my way if you like Medea, thanks.

PR #852, opened 2022-08-20
Update oc_rlvsys.lsl

Medea-Destiny:
Fix for issue #842

### DIFF
--- a/src/collar/oc_rlvsys.lsl
+++ b/src/collar/oc_rlvsys.lsl
@@ -191,8 +191,8 @@ setRlvState(){
         //Debug("RLV went inactive");
         g_iRlvActive=FALSE;
         integer relockcheck;
+        if(llListFindList(g_lBaked,["detach"])!=-1) relockcheck=TRUE;
         while (llGetListLength(g_lBaked)){
-            if(llList2String(g_lBaked,-1)=="detach") relockcheck=TRUE;
             llOwnerSay("@"+llList2String(g_lBaked,-1)+"=y"); //remove restriction
             g_lBaked=llDeleteSubList(g_lBaked,-1,-1);
         }
--- a/src/collar/oc_rlvsys.lsl
+++ b/src/collar/oc_rlvsys.lsl
@@ -4,13 +4,15 @@ Copyright (c) 2008 - 2016 Satomi Ahn, Nandana Singh, Wendy Starfall,
 Medea Destiny, littlemousy, Romka Swallowtail, Garvin Twine,
 Sumi Perl et al.
 Licensed under the GPLv2.  See LICENSE for full details.
-
 Medea Destiny   -
-        Sept 20201  -   Added RLV_CMD_OVERRIDE function. This allows one shot (=force) commands to be sent that will override any restrictions. 
-                        This command should only be used where operator has owner permission. The notion behind it is that owners should
-                        not be restricted by wearer restictions. They can unset restrictions manually, perform the function, and then reset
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
Testing required:

Proper handling of collar lock when being grabbed and released via the relay. Ensure that this does not leave an unlocked collar locked, or a locked collar unlocked, and that the checkbox in the main menu for the collar lock continues to correspond to actual status. Thus it should be ticked when the collar is locked, and unticked when the collar is unlocked, but also unticked if the collar is unlocked by an @detach=n has been issued due to relay restrictions.

pfifo:
It seems that a locked collar will remain locked when RLV is turned off and g_lBaked does NOT contain 'detach' which seems to be possible IF the code added on line 281 is triggered, because g_lBaked is not updated along with the llOwnerSay("@detatch=n") that is called.
It may be better to call ApplyAdd("detach") instead of calling llOwnerSay() directly.

Medea-Destiny:
@pfifo Good spot -- a slightly different issue, but indeed if RLV is turned off while the relay has active restraints (including detach) this would have the impact of unlocking the collar. Easy enough fix!

Medea-Destiny:
Have updated the code in question as follows:

```
        integer relockcheck;
        if(llListFindList(g_lBaked,["detach"])!=-1) relockcheck=TRUE;
        while (llGetListLength(g_lBaked)){
            llOwnerSay("@"+llList2String(g_lBaked,-1)+"=y"); //remove restriction
            g_lBaked=llDeleteSubList(g_lBaked,-1,-1);
        }
        if(relockcheck==TRUE && g_iCollarLocked==TRUE) llOwnerSay("@detach=n");
```
... so that we check the baked list for a detach, and restore the @detach=n if it is found and the collar is locked.

As for unlocking the collar if RLV is switched off -- as far as I have always understood, the RLV off function should be considered relating to functions in the RLV menu, not necessarily all RLV functions, so the collar shouldn't be unlocked by it. However if anyone has differing opinions (@SilkieSabra ?) speak up!

Pingout:
I tried to devise a test for this change, but i was somewhat confused as to the expected behavior.
First, we have spent a week, using various test situations, to check the collar lock/unlock statud checkbox. We tried teleporting, surviving a crash, surviving a sim restart etcetra. The checkbox remained in a correct status, whether locked or unlocked.
Second, I set up a BDSM device, using AV Sitter scripting which allows a command to set restrictions for the submissive on capture. Here's part of the AVpos notecard:
     RLV 1
     ROLES S
     ONTOUCH ASK
     ONSIT CAPTURE
     TIMELOCK 2
     RECAPTURE 1
     ONCAPTURE @detach=y
Using the previous oc_rlvsys script i was able to make the collar detachable even though the Lock checkbox was still in a Locked state.  With the newest script (https://github.com/OpenCollarTeam/OpenCollar/compare/Medea-Destiny-patch-3) the furniture was no longer able to defeat the collar's @detach=n
I think this means the test was successful, but i will leave it to @Medea-Destiny and @SilkieSabra to decide

SilkieSabra:
I agree that turning RLV off in the collar should not unlock the collar.  Good spot.

Pingout:
> Thus it should be ticked when the collar is locked, and unticked when the collar is unlocked, but also unticked if the collar is unlocked by an @Detach=n has been issued due to relay restrictions.

The collar lock works irrespective of RLV status

Medea-Destiny:
@Pingout Thanks for the thorough testing! Yes this is as expected and we're good to merge. :)

Issue #848, opened 2022-08-19
Addendum to Issue #844


SilkieSabra:
Reference Issue#844, Trusted is also able to alter the Exceptions list and shouldn't be.

Issue #844, opened 2022-08-17
Owned Wearer Not Self Owned should not have the ability to change the Exceptions List.


SilkieSabra:
While testing the 8.2.2 Beta we noted that an owned wearer can uncheck the Force TP exception.  We did not test the other exceptions.  This should not be the case.  Exceptions should be Owner only, otherwise a wearer has an easy way to prevent an owner from using the Force TP.  I don't know when it snuck in.

Medea-Destiny:
This should be fixed now, contact me in world for testing.

jotasprout:
I will. I'm available.

SilkieSabra:
Can you move it to a pr for testing? That way we will not have to test it twice before merge.

PR #853, opened 2022-08-20
Update oc_rlvextension.lsl

Medea-Destiny:
Fix for issues #844 & #848

### DIFF
--- a/src/collar/oc_rlvextension.lsl
+++ b/src/collar/oc_rlvextension.lsl
@@ -3,7 +3,6 @@
 This file is part of OpenCollar.
 Copyright (c) 2018 - 2019 Tashia Redrose, Silkie Sabra, lillith xue, Medea Destiny, Nirea Mercury et al.                           
 Licensed under the GPLv2.  See LICENSE for full details. 
-
 Medea Destiny -
     Sept 2021   -   Moved Exceptions menu into RLV as a main directory, and folded Menu Settings into
                     RLVSuite menu customize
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
Testing required:
Ensure that only Owners can change the exceptions list.

Pingout:
Tested by Ping and Omnia:

Ping is Owner in Rio's collar, Omnia is Trusted in that same collar
Both of us are able to change the Exceptions list for both Owner and Trusted
The changes Omnia makes are changing both Owner and Trusted exceptions, but according to the script she should be blocked from changing exceptions

I think the script fix is not working yet

Medea-Destiny:
@Pingout When you go into the exceptions menu (the one with buttons for Owner, Trusted, Custom and BACK) anyone without owner permission should see the message "No access to changing exceptions!" and be kicked back up a menu level to the RLV menu when clicking on any button other than the BACK button. Can you clarify what the point of failure is please?

Pingout:
> @Pingout When you go into the exceptions menu (the one with buttons for Owner, Trusted, Custom and BACK) anyone without owner permission should see the message "No access to changing exceptions!" and be kicked back up a menu level to the RLV menu when clicking on any button other than the BACK button. Can you clarify what the point of failure is please?

Oki, so here's what happened. Before the test I copied the new script into the test collar, and then, thinking i needed the rest of the updated files I ran the 8.2.2.010 Beta Updater. Which over-wrote the new rlvextentions with the older one.

Once we had the correct script in the collar, the script worked as it's supposed to, blocking Trusted, or any non-owner, from accessing the Exceptions list. Test successful.

@SilkieSabra  this PR can be merged.

SilkieSabra:
@Pingout if you say it can be merged, contribute the approving review to that effect.

Pingout (approved):
Approved as noted above. Thanks Medea!

PR #861, opened 2022-08-22
add feature custom root paths

tayaphidoux:
This update adds a special feature but allows oc_folders to behave as it normally does when the options are not used.
changes:
    1) folders can be defined as root with a note card list, each line of the note card can be used to define a path and button that will be added to the
    main menu buttons, and enable the ability to disable browse button under config defined as "Default root".
   2) disabling "Default root" removes browse from the menu and allows only the paths defined by the note card to be used.
   3) folders with (~) or (.)will have the leading character striped from the button label but still function.
   4) folders will have the first letter capitalzied for the button label for readability. 
   5) not tested you can define full folder paths this allows for folders nested inside other folders.

### DIFF
--- a/src/collar/oc_folders.lsl
+++ b/src/collar/oc_folders.lsl
@@ -12,6 +12,9 @@ Medea (Medea Destiny)
                         term, added sanity checking for chat commands to ensure we don't try to search for nothing,
                         and no longer operate on empty search results. Issuer of chat command now stored as 
                         g_kChatUser so they can be notified if findfolder fails.
+Phidoux (taya.Maaruti)
+    * Aug 2022      -   * Add the ability to set custom root folders using a notecard and disable Browse button
+                        effectivly removing #RLV as root folder for browsing.
                          
 et al.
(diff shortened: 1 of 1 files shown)


SilkieSabra:
This looks good as best I understand it, it's a step beyond the last bug fixing we are doing for 8.2.2.

tayaphidoux:
@SilkieSabra  it is but its one i could not wait to introduce it actually covers a question i was asked a while back, and i felt it had so much potential i wanted to share it even if it don't make it into 8.2.2, the person was asking about a way to hide wardrobe related folder from # folders, this would allow them to do that by defining a different directory as root while disabling the old browse button, in turn this would hide any sub-folders that don't assume ~ or . to hide them from systems like this.

tayaphidoux:
noticed a checkbox bug in current version will have to find out what causes it and fix it.

tayaphidoux:
closing to wait and see what happens and there is a  bug with first button showing check when other items are worn else where.

Issue #859, opened 2022-08-22
Found another Wearer as Trusted conflict -- Unwelder


SilkieSabra:
See #844 -- When Wearer is also Trusted on the Collar, they are denied access to use the unwelder.  Remove them from the Trusted list on their own collar and they can use the unwelder.

PR #862, opened 2022-08-24
Update oc_unwelder.lsl

Medea-Destiny:
Fix for #859


Medea-Destiny:
Okay. So I think this should do it. To test:
Wearer with trusted permission can now unweld as normal.
Owner of unwelder cannot unweld collars they do not have permissions for (i.e not wearer or owner).
Both unwelder and unwelder HUD are working as intended.

Pingout:
Omnia and I tested this: 
1. Wearer with trusted permission can now unweld as normal.
2. The second test is not valid: There's no way for someone who is not the collar wearer to unweld a collar
3. Both unwelder and unwelder HUD are working as intended.
Thank you for the fix!

SilkieSabra:
I'm being a stickler here for checking that Approved box.

Pingout (approved):
Approved as noted above.

Pingout (approved):
Approved as noted above.

PR #865, opened 2022-09-21
Yosty fixes - remove Setor() and Bool()

Medea-Destiny:
Removes SetOr() and Bool() functions from OC core to streamline code - saves about 2KB of script RAM.

This PR replaces #702 to make merging easier, but is the same code. Full credit to @Yosty7b3


Pingout (approved):
Tested by Ping and Omnia, these all look great. Thanks Yosty and Medea

PR #866, opened 2022-09-21
Yosty Fixes - Update oc_api.lsl

Medea-Destiny:
Changes to reset oc_api only when needed rather than every boot event. Part of speeding up boot process. This PR replaces #733  but is the same code. Full credit to @Yosty7B3


Pingout (approved):
Awesome, thank you.

PR #874, opened 2022-10-20
update control method 2

tayaphidoux:
This uses the method i mentioned here will need next version.txt updates to be 8.2.3.000 or what ever it will be. https://github.com/OpenCollarTeam/OpenCollar/pull/873#issuecomment-1284633512


tayaphidoux:
i guess these should be closed since you are just automatically going to go with media's version without a discussion.

PR #882, opened 2022-10-21
fix a few issues roan found

tayaphidoux:
Fixed the issue of note card trying to load non available one, if its not available and the user clicks they will be prompted to select. Fixed the issue with llSetOverride persistance when removed while active. Fixed the issue with Permissions failing due to owner key change.

### DIFF
--- a/src/ao/oc_ao.lsl
+++ b/src/ao/oc_ao.lsl
@@ -3,7 +3,11 @@
 // littlemousy, Romka Swallowtail, Garvin Twine et al.
 // Licensed under the GPLv2.  See LICENSE for full details.
 /*-Authors Attribution-
-Taya Maruti - (May 2021)
+Taya Maruti
+    (May 2021) - combined oc_addon updates and split oc_ao_animator to reduce stack heap collisions and improve script time.
+    oct 20 2022 - fixed the load default card issue when default not available.
+                    fixed permission error when passed from one user to the other.
+                    fixed reset on removal without turning off.
 */
 
 integer API_CHANNEL = 0x60b97b5e;
@@ -573,7 +577,6 @@ default {
     }
 
     on_rez(integer iStart) {
-        if (g_kWearer != llGetOwner()) llResetScript();
         if (g_iLocked) llOwnerSay("@detach=n");
         g_iReady = FALSE;
         llMessageLinked(LINK_THIS,AO_SETTINGS,"UPDATE",g_kWearer);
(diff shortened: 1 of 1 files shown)


SilkieSabra (changes requested):
I am testing this inworld based on the compiled scripts Phidoux sent me, on a newly updated 8.2.2 collar.  It is a modified collar but the only script that is different from standard is oc_titler plus a few additional scripts not related to oc_anim or the AO.  
None of the animation sets are loading.  I do not see any of the animations running in Animation Explorer or visibly.  Three of the sets report they were loaded successfully yet do not run:  Girl, Boy, and Pony.  The other two sets, Crawl and Worm, delivered the same script error: Could not find animation '﷐'"".

Omnia2021:
Testing the new AO 2.2 scripts and have had a variety of issues. After attaching it it fails to recognize some of the animation sets. Loading times are longer than you would expect between 15 and 30 seconds per,and some messages are contradictory to what occurs, example while loading animation set Crawl I got an invalid animation set message but after a few seconds it activated. The issues seem to be random, Pony doesn't work at all after saying it loaded successfully and boy & girl work but take 20 plus seconds to load. Pony loads but then an error message pops up "Could not find animation". Worm immediately produces the "could not find animation" message. Testing at Moses Test Site, with fully updated collar. These issues seem to match Roan's experience.

tayaphidoux:
> I am testing this inworld based on the compiled scripts Phidoux sent me, on a newly updated 8.2.2 collar. It is a modified collar but the only script that is different from standard is oc_titler plus a few additional scripts not related to oc_anim or the AO. None of the animation sets are loading. I do not see any of the animations running in Animation Explorer or visibly. Three of the sets report they were loaded successfully yet do not run: Girl, Boy, and Pony. The other two sets, Crawl and Worm, delivered the same script error: Could not find animation '�'"". AO will have to be working before I can deliver the update to the vendors and complete the full release, prior to opening a new branch for the next release.

this is the ao script you need both oc_ao and oc_ao_animator

tayaphidoux:
> Testing the new AO 2.2 scripts and have had a variety of issues. After attaching it it fails to recognize some of the animation sets. Loading times are longer than you would expect between 15 and 30 seconds per,and some messages are contradictory to what occurs, example while loading animation set Crawl I got an invalid animation set message but after a few seconds it activated. The issues seem to be random, Pony doesn't work at all after saying it loaded successfully and boy & girl work but take 20 plus seconds to load. Pony loads but then an error message pops up "Could not find animation". Worm immediately produces the "could not find animation" message. Testing at Moses Test Site, with fully updated collar. These issues seem to match Roan's experience.

on the scripts i just uploaded here both of them loading times are near instant with default ao as well as i have seen no issues

tayaphidoux:
nobody complained of these issues when it was first tested.

tayaphidoux:
https://github.com/OpenCollarTeam/OpenCollar/pull/883 - you need this along side the script in this pr

tayaphidoux:
@Omnia2021 @SilkieSabra 
Make sure the characters in the note card are not special characters and match the animation in inventory exactly else you will get missing animation because names don't match.

PR #883, opened 2022-10-21
Fix llSetOverride Persistance issue in oc_ao_animator

tayaphidoux:
fixing the llOveride Persistance issue when detaching the ao while its active.


tayaphidoux:
the other script was closed and they are out alignment with the project now also i have been working on a linkset data replacement

PR #885, opened 2022-10-29
8.2.3

Medea-Destiny:
Various tweaks to update to 8.2.3 emergency release

What happened? We released 8.2.2 with 8.2.1 scripts in it. Rather than re-releasing with the same version number, which has the potential to cause confusion, we need to bump build to 8.2.3.  This PR is to bring master in line with the updates I've done inworld to the updater to make 8.2.3. It contains a small number of minor modifications where we've omitted a change to internal script numbers for debug, or haven't added in changelog notices. There's also a version check tweak, for which see below.

 oc_addons: bump version number, add omitted credits
oc_anim: add omitted credits
oc_bookmarks: bump version number
oc_core: bump version number, fix for versioning issue (see below)
oc_rlvsys: bump version number, change "RLV ready!" message to "RLV active!" to avoid confusion in boot process (people sometimes thought it meant the collar was ready)

### DIFF
--- a/src/collar/oc_addons.lsl
+++ b/src/collar/oc_addons.lsl
@@ -9,6 +9,14 @@ Medea (medea.destiny)
     Sept 2021           -   Provided auth failure mode for menu. Insufficient auth now
                             provides suitable notification to user and respawns main menu. 
                             Fixes issue #665   
+
+KristenMynx
+    Dec 2021            -   Fix timeout removal stride
+    May 2022            -   Fix offline removal stride
+                        -   Reduce chatter
+    Jun 2022            -   Reduce chatter more
+    
+    
     
 et al.
 Licensed under the GPLv2. See LICENSE for full details.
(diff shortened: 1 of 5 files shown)


Medea-Destiny:
Fix for versioning (see #873 #874).

Fix is below: 

```
else if(iV1 > iV2){
        if(llGetSubString(V2,-3,-1)=="000" && llGetSubString(V1,0,-4)==llGetSubString(V2,0,-4)){
            UPDATE_AVAILABLE=TRUE;
            g_iAmNewer=FALSE;
            return;
        }
```
explanation: Where v1 is greater than v2 (for example, current version v1 is 8.2.3000 and new version v2 is 8.2.3030) we check to see whether the last three digits of the "new" version are "000" and if the rest of the strings are the same. In this case, we know that the current version is less than the new version only because of an RC, beta or alpha code, which 000 should supersede.

This method doesn't require changes to version format, is quick and easy, and will change behaviour only in the case where the updater version is 000 (i.e. a full release version), in which case it will count as newer than an RC, Beta or alpha build. This method does not, as the question was raised earlier, take into account situations where we want to release say beta 12, as there's only one digit per rc/beta/alpha code. I don't think this is an issue because when do we ever release that many betas. However if a change was made to cope with this, we'd have to dot separate every part of the version string, and fix the code which lists version numbers in the menu as well.  

This method also works for version comparisons of dev versions, which the method suggested in #874 does not.

Note that in the in-world updater I've made, this particular bit of code is currently commented out pending final decision.

PR #887, opened 2022-10-29
Installer update quick maybe works who knows thing

Medea-Destiny:
Look, I've just done this in a few minutes without properly poking around because Roan was DESPERATE. I think this is gonna change the "Skip" button to "Keep" in cases where it's asking to remove or retain an optional script that's already in the collar, but I haven't tested this at all and I only had a few minutes to poke at the installer so there has been no proper logic testing or ANYTHING.

So basically I think it's gonna work, but it might not. I dunno! It's here because Roan wanted it fast. I've done it fast. It might work. Try it out and see! Yay fun. :P

### DIFF
--- a/src/installer/oc_installer_bundles.lsl
+++ b/src/installer/oc_installer_bundles.lsl
@@ -140,7 +140,7 @@ default
             g_lTypes = llDeleteSubList(g_lTypes, iIndexChannels, iIndexChannels);
 
             llListenRemove(iLstn);
-            if(sMsg == "Skip"){
+            if(sMsg == "Skip" ||sMsg=="Keep"){
                 g_iLine++;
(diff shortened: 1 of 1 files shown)


SilkieSabra:
I'm not going to wait the update for this to be tested, I will wait making the next version branch for it though.  Since it's not part of the collar itself and it's just a change in wording and won't affect anything downstream,  we can merge this and sneak the new one into the vendors ahead of the next release.

Pingout (approved):
Yup, this works great, thank you!

SilkieSabra (commented):
The review requirement got re-opened I think because a web text changed in the master branch, and i merged master back to this branch after it was reviewed.  I'm going to go ahead and override to merge.

Issue #892, opened 2022-11-02
Cuff Themes Bug


Pingout:
What version of OpenCollar are you using? Collar 8.2.3, Cuffs 2.2 

What behavior did you expect? There is a routine in the Cuff menu to create a new theme, as long as you have full permission

What behavior did you see instead? The New Theme menu button does nothing - oc_cuff_themes.lsl script bug

This is pretty low priority. Most casual users would never even see the New Theme button in their cuff menu. This is something only Collar/Cuff designers would use, and in most cases a quick cut & paste from the Collar Theme will do the trick. But if we're going to have a New Theme button in the menu, it should function properly. We somehow missed this bug over the course of six months of testing, and 3 separate Beta releases.

Pingout:
Merged and closed

PR #893, opened 2022-11-02
Update oc_cuff_themes per Issue #892

Pingout:
It looks like there was a single number incorrect in the oc_cuff_themes script. The oc_cuff script was sending iNum == 33, and the oc_cuff_themes script was looking for iNum == 32


Pingout:
This update also has a problem. When you have multiple cuffs, all of them get the message to dump parameters to PrintCurrentProperties, and you end up with a huge mixed up mess in Local Chat

Pingout:
Closing this, I will rewrite it for clarity

PR #894, opened 2022-11-03
Changes to oc_cuff for New Theme functionality

Pingout:
Proposed fix for Issue #892

### DIFF
--- a/src/cuffs/oc_cuff.lsl
+++ b/src/cuffs/oc_cuff.lsl
@@ -15,6 +15,8 @@ Kristen Mynx
     * May 2022        -       Added resizer support
 Kristen Mynx
     * July 2022       -       Fix "BACK" buttons on resizer 
+Ping (Pingout Duffield)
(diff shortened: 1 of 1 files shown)


Omnia2021:
Tested the updated New Theme script with updated scripts in full perm OpenCollar Addon cuffs. Script behavior was as expected with the only concern being the generous amount of output messages in chat. Parameters were present only with cuffs that were clicked, as it should be. Theme button was present in cuffs with poses only, as it should be.

Omnia2021 (approved):
Tested the updated New Theme script with updated scripts in full perm OpenCollar Addon cuffs. Script behavior was as expected with the only concern being the generous amount of output messages in chat. Parameters were present only with cuffs that were clicked, as it should be. Theme button was present in cuffs with poses only, as it should be.

SilkieSabra:
I'm overriding the Review required by a reviewer with write access, since we do have one approving review.  I'm assuming by "New Theme Script" you mean the "New Theme" feature in the menu that allows creators to set a new theme for the cuffs.

PR #901, opened 2022-11-16
This is the 6 panel Plane for MeshLabel

Pingout:
This is best uploaded using Cube Physics to leave enough size for collar resizer. Number consecutively as shown in the script. Texture, Offset, and Scale are set in the oc_meshlabel script. Recommend using the Full Bright option.


Pingout:
@SilkieSabra Do you think we should update the Complete User Guide collar build document to include the instructions from the oc_meshlabel script:

// How to use:
// 
// - You need a Mesh-Strip with 6 Faces next to each other. Each Face should have 1 Material.
// - Link the Strips together next to each other. 
// - Name them like this: MeshLabel~number~line
//      number is the mesh-number from 0 to how many you linked horizontally   
//      line is the vertical number from 0 to how many lines you have

The Stock OpenCollar has 4 of these objects, named MeshLabel~00 thru MeshLabel~03

Issue #907, opened 2022-12-06
Suggestion: unwelder check order of welded <-> addonslimit


Pyhrn:
Currently when starting the unwelding-process the script checks first for the addonslimit flag, and only after that if the collar is even welded. 
Imho it is rather pointless to send the wearer on _an adventure_ to their owner for lifting addonslimit just to find out on return their collar hasn't been welded in first place.

PR #911, opened 2022-12-10
Unwelder: Possible solution to Issue #907

Pingout:
Slight change to the order of operations to first check if the collar is welded before checking if the Addon Limited is unchecked


SilkieSabra (approved):
Tested this at the HQ with Yshida, it performs as expected.

SilkieSabra:
I can merge this as soon as you move it from master to 8/3_Features_branch.

SilkieSabra (approved):
This has been working well inworld for a week or so now.

PR #917, opened 2022-12-11
Update oc_unwelder_hud.lsl

Pingout:
Fix for Issue #907, order of welded <-> addonslimit checking


SilkieSabra:
We just need somebody else besides Ping to test the HUD version.  Not broken; checks for existing weld before checking for AddOns Limits on or off.

PR #919, opened 2022-12-12
oc_teleport HUD

Sander8787:
this is the code that goes into the HUD that interacts with the oc_teleport's Follow Me feature.

this code could later be incorporated into the OpenCollar remote HUD, since it is a feature that would be used by the collar owner to interact with the collar wearer.

### DIFF
--- a/teleporterHUD
+++ b/teleporterHUD
@@ -1,0 +1,94 @@
+integer Active;
+
+key regionQuery = NULL_KEY;
+
+key http_request_id;
+string url = "https://sldb.app/";
+
+default
+{
+    attach(key id)
+    {
+        if(id)  llResetScript();
+    }
+    state_entry()
+    {
+        //to override NO script zones//
+        llRequestPermissions( llGetOwner(),PERMISSION_TAKE_CONTROLS);
+         
+        //display ON//
+        llSetLinkColor(LINK_ALL_OTHERS, <0,0.6,0>, ALL_SIDES);
+        llSetText("ON", <1,1,1>, 1.0);
+        Active = TRUE;
+        
+        //get location data//
+        regionQuery=llRequestSimulatorData( llGetRegionName(), DATA_SIM_POS );
+    }
+
+    touch_start(integer total_number)
+    {
+        if( Active )
+        {
+            //display OFF//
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
Hi Sander, oc_teleport is not a script that is part of the official repository, we're going to need some more information about this.

Sander8787:
ja, I was told to list it on the github if I wanted to recommend it be integrated in the collar.

Its an App I made. It has 2 features:
1-> SimLock: locks collar wearer to a Sim, prevent to TP away.
2-> Follow Me: follows collar wearer to owner, everything the owner TP somewhere new, collar wearer TP to that location also!

Pingout:
> ja, I was told to list it on the github if I wanted to recommend it be integrated in the collar.
> 
> Its an App I made. It has 2 features: 1-> SimLock: locks collar wearer to a Sim, prevent to TP away. 2-> Follow Me: follows collar wearer to owner, everything the owner TP somewhere new, collar wearer TP to that location also!

Hi Sander: I only see one part on the github. There should be two, as you said: One script for the collar apps, and one for the Owner HUD.  Thanks

Sander8787:
HUD script:
https://github.com/OpenCollarTeam/OpenCollar/compare/master...Sander8787:OpenCollar:patch-2
oc_teleport:
https://github.com/OpenCollarTeam/OpenCollar/compare/master...Sander8787:OpenCollar:patch-1

SilkieSabra:
Is this a feature OpenCollar could use? It's been waiting around in the PRs for a long time.

mistressohm:
Perhaps an improved leash holder could use this type of remote-access
teleport functionality.  On teleport events, store the UUID of the leashed
collar wearer, tell the collar to Unleash, then when teleport completes,
send an automatic tplure to the stored UUID and re-leash once they're
nearby again.

On teleport crash (teleport never completes, instead the leash holder is
disconnected), the collar is already un-leashed, so no more being yanked
halfway across a sim when the leash holder relogs back into the wearer's
region.

Potential issues - if the wearer is leashed to a non-owner, the tplure
might fail or the wearer may have option to refuse.  But that's also a
safety feature for capture play.  The leash holder can't just port to
another sim and drag the 'victim' with them unless their collar is set to
allow no-prompt teleports from non-owners.

PR #924, opened 2023-01-02
Update oc_resizer.lsl

Pingout:
Need to fix the behavior of thr BACK button in the Position Menu, Rotation Menu and Size Menu


Omnia2021 (approved):
Fix worked as expected with no issues, much more user friendly now.

Issue #932, opened 2023-01-25
Meshlabel script bug


Pingout:
What version of OpenCollar are you using? Current Version 8.2.3

What behavior did you expect? Meshlabel should hide when the collar is hidden

What behavior did you see instead? Meshlable stays visible

What steps does someone need to take to reproduce the problem? The problem is in line #191 of the oc_meshlabel script, setting incorrect primitive params.

PR #933, opened 2023-01-25
Update oc_meshlabel.lsl

Pingout:
Ref Issue #932

### DIFF
--- a/src/collar/oc_meshlabel.lsl
+++ b/src/collar/oc_meshlabel.lsl
@@ -5,7 +5,9 @@
 // Joy Stipe, Wendy Starfall, Romka Swallowtail, littlemousy,       
 // Garvin Twine et al.   
 // Licensed under the GPLv2.  See LICENSE for full details. 
-
+//
+// Ping (Pingout Duffield) 
+//         *Aug 2023 - Revise visibility when collar is hidden Ref Issue #932
 // How to use:
 // 
 // - You need a Mesh-Strip with 6 Faces next to each other. Each Face should have 1 Material.
--- a/src/collar/oc_meshlabel.lsl
+++ b/src/collar/oc_meshlabel.lsl
@@ -189,7 +189,7 @@ integer LabelsCount() {
             lLineLinks += [0]; // Fill with Zero
             g_lLabelLinks = llListReplaceList(g_lLabelLinks,[llDumpList2String(lLineLinks,"|")],iLine,iLine);
             //change prim description
-            llSetLinkPrimitiveParamsFast(iLink,[PRIM_DESC,"Label~notexture~nocolor~nohide~noshiny"]);
+            llSetLinkPrimitiveParamsFast(iLink,[PRIM_DESC,"Label~notexture"]);
         } else if (sLabel == "LabelBase") g_lLabelBaseElements += iLink;
     }
     


Medea-Destiny (changes requested):
Okay SO... the reason why the nohide tag is set is that the label scripts can turn the label on and off themselves. If we simply remove the nohide tag as per this pull, then if you switch the label off and then hide the collar, these prims will get hidden. When you show the collar again, these prims will then be set opaque and show, despite the label supposedly being hidden.

What needs to be done for the label scripts is to keep the nohide tag, but check for the global_hide variable being set. So in the LM_SETTINGS_RESPONSE section of the link message event, where we check for globals (line 514) add another check after the checkboxes value,
 ```
if(sToken=="hide") { 
      if(sValue=="1") g_iHide==TRUE;
      else g_iHide==FALSE;
      SetLabelBaseAlpha();
}
```

Then we need to fix the SetLabelBaseAlpha() function by deleting the if(g_iHide) return; and making it set the alpha to transparent if either g_iHide is true OR g_iShow is false, or opaque if g_IHide is false  AND g_iShow is true. So before the for loop, define:
```
 integer iVal=0; 
 if (g_iHide==FALSE&&g_iShow==TRUE) iVal==1; 

```
Then replace the g_iShow inside the for loop ( lines 262 and 263) with iVal.

Finally, remove the if (iChange & CHANGED_COLOR)  section in the changed event. This is attempting to set the g_iHide based on a change event being triggered by the collar going transparent, however it'll break if the root prim is always transparent, and clashes with the if(g_iHide) return inSetLabelBaseAlpha();

I haven't tested any of this of course but I think that'll do it.

Pingout:
I was able, I think, to put the changes suggested into a clean version of the oc_meshlabel script, and it compiled fine in SL. However, the changes didn't have the desired effect. The Meshlable object did not turn alpha with the rest of the collar when using the Collar Hide command.
I did a little more testing with the edited script here, the script with changes to the line that sets the Meshlabel object description. I did not see the effect Medea though would happen, where the Meshlabel turns opaque and can be seen. I believe the oc_meshlabel script controls the object texture, forcing a transparent texture at all times. I tried a half dozen different ways, trying to force the collar into the effect of a visible opaque label, and none would work, the collar always displayed the transparent prims

SilkieSabra:
Why do we need the label and meshlabel to show/hide independently of the collar? I can't think of any purpose for that and suspect it's been broken for years.  Just take out that "feature".

Pingout:
I revisited this script bug fix and I stand by my comments from February 2. 
1) The bug fix suggested by Medea didn't have the desired effect; that of turning the MeshLabel object transparent with the collar Hide. 
2) I also did not see the effect Medea was concerned about, that of having the MeshLabel Object turn opaque (visible).
3) The initial problem was the MeshLabel Object *not* turning invisible when using the Hide command in the collar. If the Label is in use, ie. words or numbers on the meshlabel object, the words show up on the wearer's neck when the collar is hidden. The wearer would need to open the Label App, and turn off the text by unticking the Show checkbox.
4) By changing the coded primitive parameter to Description: Label~notexture (removing the forced ~nohide) the meshlabel does what it is supposed to do, that being Hide when the rest of the collar is Hidden. 
5) It's not a question of having the Label Show/Hide independently of the collar. That feature is managed by the oc_meshlabel script, and works as it should. The meshlabel object only displays the label text when a line is set in the menu, and the Show checkbox is ticked.
6) I initially tested the script using the MeshLabel Object created by Kyrah Abattoir. I also created a MeshLabel Object in Blender, importing it to LS. Both versions of the MeshLabel Object work as expected. I uploaded this MeshLabel Object to the Github at https://github.com/OpenCollarTeam/OpenCollar/blob/master/res/models/MeshLabel%20Object.dae

SilkieSabra:
This needs @Medea-Destiny to take another look and clarify, and resolve this so we can move on the pr.

Medea-Destiny (changes requested):
Yeah this is fine.

 So, on closer inspection the setlabel function is called after anything happens in UserCommand, so after hiding the mesh label prims by setting the alpha, it then also overwrites them with spaces. When you use hide then show in the collar settings menu it will indeed show the mesh label prims as I assumed; however because they're showing spaces if they've been hidden, you don't see anything. Setting the mesh label faces to transparent is actually pointless given they're set to spaces immediately afterwards anyway. Ugh.

One small change needed, @Pingout : credit note/date at the top with brief description of the change and ref to issue #932. Thanks!

Omnia2021 (approved):
Works as intended.

PR #941, opened 2023-03-09
Linkset Data Preperation

tayaphidoux:
this update incorporates a little linkset data which will help to prepare collars for the future.

### DIFF
--- a/src/collar/oc_settings.lsl
+++ b/src/collar/oc_settings.lsl
@@ -10,6 +10,9 @@ https://github.com/OpenCollarTeam/OpenCollar
 Medea Destiny
         Sept 2021   -   Tightened timings on release of all settings and reduced sleep
                         padding on initialize
+
+Taya Maruti
+        Mar 2023    -   Preparation for linkset data migration.
 */
 
 //string g_sParentMenu = "Apps";
@@ -20,13 +23,22 @@ integer TIMEOUT_FIRED = 30499;
 
 
 string GetSetting(string sToken) {
-    integer i = llListFindList(g_lSettings, [llToLower(sToken)]);
-    if(i == -1)return "NOT_FOUND";
-    return llList2String(g_lSettings, i + 1);
+    sToken = llToLower(sToken);
+    integer i = llListFindList(g_lSettings, [sToken]);
+    if(i == -1)
+    {
+        return llList2String(g_lSettings, i + 1);
+    }
+    else if(~llListFindList(llLinksetDataListKeys(0,llLinksetDataCountKeys()),[sToken]))
+    {
+        return llLinksetDataRead(sToken);
+    }
+    return "NOT_FOUND";
 }
 
 DelSetting(string sToken) { // we'll only ever delete user settings
     sToken = llToLower(sToken);
+    llLinksetDataDelete(sToken);
     integer i = llGetListLength(g_lSettings) - 1;
     if(sToken == "intern_weld")DeleteWeldFlag();
     if (SplitToken(sToken, 1) == "all") {
@@ -64,12 +76,22 @@ integer GroupIndex(string sToken) {
 
 integer SettingExists(string sToken) {
     sToken = llToLower(sToken);
-    if (~llListFindList(g_lSettings, [sToken])) return TRUE;
+    if(~llListFindList(llLinksetDataListKeys(0,llLinksetDataCountKeys()),[sToken]))
+    {
+        if(~llListFindList(g_lSettings, [sToken]))
+        {   // reapir g_lSettings if missing.
+            g_lSettings = SetSetting(sToken,llLinksetDataRead(sToken));
+        }
+        return TRUE;
+    }
+    else if (~llListFindList(g_lSettings, [sToken])) return TRUE;
     return FALSE;
 }
 
 list SetSetting(string sToken, string sValue) {
     sToken = llToLower(sToken);
+    // write every thing to linksetdata
+    llLinksetDataWrite(sToken,sValue);
     integer idx = llListFindList(g_lSettings, [sToken]);
     if (~idx) return llListReplaceList(g_lSettings, [sValue], idx + 1, idx + 1);
     idx = GroupIndex(sToken);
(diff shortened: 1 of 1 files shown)


Pingout (approved):
We've been testing this little App for a few days, so far everything is going great. I like that the collar Owner can also retrieve the linkset data from the collar.

Medea-Destiny:
I think it's a mistake to attempt to incorporate LSD in this fashion, maintaining it alongside settings. To benefit properly from LSD we need to recognize and account for the fact that it doesn't need to be centrally managed, and can be edited without going through the settings script. As this stands we could potentially have a setting stored both in g_lSettings and as LSD which are out of sync, which could cause real problems. 

We need to make this LSD transition for settings in a single step, ensuring that we have something robust in place - a fairly fundamental rewrite of oc_settings is the transition step, but we can utilise that to make the transition across the collar as a whole more graceful. We need to interpret the current link_message method for settings, so that the change is transparent to scripts that are not yet updated to use LSD, but we should be utilizing LSD internally in all those cases rather than attempting to maintain g_lSetting alongside LSD. As part of the transition process we also need to consider the update process, which means that the new oc_settings script must be able to hand the conversion of settings sent by an updater from 8.x and older versions.

Other considerations are namespaces and memory management. Is the current settings namespace robust for LSD? I'm not sure it is, as we're currently storing all global settings centrally and with LSD both global and local settings are going to be stored, so we'll need to ensure it's easy to identify global LSD settings, and may need to rethink token prefixes slightly. There's also the question of managing the usage of LSD memory to consider -- are we going to reserve space and monitor memory usage? If so this would be a task that oc_settings should be set up to handle.

Medea-Destiny (changes requested):
I would strongly advise against a merge at this stage -- we need a more fundamental rewrite of oc_settings rather than a half-way house. We can't transition the collar over in one go, but oc_settings is the script that will handle that transition and we need to have a robust definition for OC's LSD usage in place, and we need to replace g_lsettings entirely and have link_message handling in OC_settings for backwards compatibility in place as a temporary measure while the rest of the transition takes place. See my comments in conversation for more details.

tayaphidoux:
@Medea-Destiny 
problem one can be solved by listening for changes using linkset_data(integer iAction,string sName, string sValue){} and setting them to gl_settings, making linkset data more prioritized if that's your concern, i have this script in my collar right this moment and its been working without issue.

one of the advantages of linkset is to reduce script events which reduces the cost of a script making it more efficent and lighter on the sims, having a bunch of llLinkMessages() is one of the problems with open collar right now.

look at my ao work if you want to see i can send you my current ao which is 90% linkset storage.

> I think it's a mistake to attempt to incorporate LSD in this fashion, maintaining it alongside settings. To benefit properly from LSD we need to recognize and account for the fact that it doesn't need to be centrally managed, and can be edited without going through the settings script. As this stands we could potentially have a setting stored both in g_lSettings and as LSD which are out of sync, which could cause real problems.
> 
> We need to make this LSD transition for settings in a single step, ensuring that we have something robust in place - a fairly fundamental rewrite of oc_settings is the transition step, but we can utilise that to make the transition across the collar as a whole more graceful. We need to interpret the current link_message method for settings, so that the change is transparent to scripts that are not yet updated to use LSD, but we should be utilizing LSD internally in all those cases rather than attempting to maintain g_lSetting alongside LSD. As part of the transition process we also need to consider the update process, which means that the new oc_settings script must be able to hand the conversion of settings sent by an updater from 8.x and older versions.
> 
> Other considerations are namespaces and memory management. Is the current settings namespace robust for LSD? I'm not sure it is, as we're currently storing all global settings centrally and with LSD both global and local settings are going to be stored, so we'll need to ensure it's easy to identify global LSD settings, and may need to rethink token prefixes slightly. There's also the question of managing the usage of LSD memory to consider -- are we going to reserve space and monitor memory usage? If so this would be a task that oc_settings should be set up to handle.

tayaphidoux:
the objective of this script is step 1 to migration mirror g_lSettings and put emphasis on linksetdata over g_lSettings, the next phase would be to remove or minimize g_lSettings where possible and focouse on linkset data,
while its mirrored that is the phase to start targeting the other scripts to linkset data it allows a smoother upgrade over time rather then trying for an all at once.

tayaphidoux:
this is my collar with this oc_settings mirrored to linkset
```
[12:57]  TE - Paws Collar Safra's Pony - chest - test: Linkset Data Contains:
auth_block=bda910ab-af92-4661-acb2-3c4b64cd6e47
auth_limitrange=0
auth_owner=1460251a-0abe-4507-8585-fce6a1e76559
auth_trust=2f18e863-da58-4fbc-93cf-c3a194079258,a6c9d8ee-8e02-44db-9d60-c3f9e39f354d
bell_on=0
bell_sound=4
capture_status=8
folders_accessflags=11
folders_diaperbags=1
folders_restraints=1
folders_roots=0
folders_toolboxes=1
folders_vanities=1
folders_wardrobes=1
folders_wdistance=13.000000
global_checkboxes=▢,▣
global_debugger=1
global_hide=0
global_locked=1
global_prefix=tm
intern_weld=1
label_show=1
leash_leashlength=1
leash_turn=1
offset_hovers=back,0.540000,belly,0.120000,bellyrub,0.040000,bracelets,0.040000,cutie,-0.040000,doggie,-0.120000,rise,-0.120000,roll,0.060000,shy,-0.080000,submit,-0.040000,table,-0.120000,tower,-0.100000,beg,0.220000,nadu,0.140000
outfits_accessflags=25
particle_c_texture=Chain
particle_color=<1.000000, 0.505880, 0.623530>
particle_particlemode=Ribbon
particle_rtexture=Silk
presets=Anthro~<0.08156, 0.07141, 0.07141>/<-0.00446, 0.00018, 0
```

`[12:58]  TE - Paws Collar Safra's Pony - chest - test: Linkset Data Memory
Used/Free/Max
1540/63996/65536`

Medea-Destiny:
@tayaphidoux Yes we could use the linkset data event to monitor changes other scripts do to LSD tokens and update based on that, but why would we want to? As you say we're trying to avoid unnecessary events. There's no advantage to storing g_lSettings when we have LSD, only disadvantages. 

With LSD, the oc_settings script becomes fundamentally different in purpose to what it has been before. It's no longer necessary to have any central storage of data - this becomes counter-productive. In theory would wouldn't need this script any more at all. In practice what we're going to need the settings script for is not to handle settings storage, but simply to manage it. It'll be there for backwards compatibility at least in the medium term. It should manage storage such that for example it will handle clearing LSD data on a complete script reset. It should contain a method for data transport with a serialisation function. Etc.

tayaphidoux:
@Medea-Destiny 
step 1) (this pull request) g_lSettings  - to -> LSD
step 2) stop using llMessageLinked() in updated 
step 3) migrage old code to stop using llMessage Linked()
step 4) remove this script or any parts not still being used.

tayaphidoux:
i need to re think the section 
what i currently have will likely break things. if i knew exactly how the all function worked, and it super seeds the current unweld call.
DelSetting(string sToken) { // we'll only ever delete user settings

PR #943, opened 2023-04-02
adding LinksetData to oc_core for dynamic menus

Yosty7B3:
* All the settings are now read directly from the LinksetData instead of relying on link messages.
* Buttons can now be added and removed by directly editing the LinksetData OR through the old method of link_messages.
* The LinksetData keys used for the menu's are 'Menu_' followed by the name of the menu. So for Main that is "Menu_Main", and for Apps that is "Menu_Apps".
* The values are comma separated strings of the buttons.
* Weld has been removed mostly from oc_core in preparation of a separate oc_weld script.

### DIFF
--- a/src/collar/oc_core.lsl
+++ b/src/collar/oc_core.lsl
@@ -47,6 +47,8 @@ integer NOTIFY_OWNERS=1003;
 
 //string g_sParentMenu = "";
 string g_sSubMenu = "Main";
+string g_sRespring;
+
 string COLLAR_VERSION = "8.2.3000"; // Provide enough room
 // LEGEND: Major.Minor.Build RC Beta Alpha
 integer UPDATE_AVAILABLE=FALSE;
@@ -79,9 +81,9 @@ integer REBOOT = -1000;
 integer LM_SETTING_SAVE = 2000;//scripts send messages on this channel to have settings saved
 //str must be in form of "token=value"
 //integer LM_SETTING_REQUEST = 2001;//when startup, scripts send requests for settings on this channel
-integer LM_SETTING_RESPONSE = 2002;//the settings script sends responses on this channel
+//integer LM_SETTING_RESPONSE = 2002;//the settings script sends responses on this channel
 integer LM_SETTING_DELETE = 2003;//delete token from settings
-integer LM_SETTING_EMPTY = 2004;//sent when a token has no value
+//integer LM_SETTING_EMPTY = 2004;//sent when a token has no value
 
 integer MENUNAME_REQUEST = 3000;
 integer MENUNAME_RESPONSE = 3001;
@@ -111,9 +113,6 @@ string g_sUnlockSound="82fa6d06-b494-f97c-2908-84009380c8d1";
 
 integer g_iListenPublic=TRUE;
 
-key g_kWeldBy;
-list g_lMainMenu=["Apps", "Access", "Settings", "Help/About"];
-
 Dialog(key kID, string sPrompt, list lChoices, list lUtilityButtons, integer iPage, integer iAuth, string sName) {
     key kMenuID = llGenerateKey();
     llMessageLinked(LINK_SET, DIALOG, (string)kID + "|" + sPrompt + "|" + (string)iPage + "|" + llDumpList2String(lChoices, "`") + "|" + llDumpList2String(lUtilityButtons, "`") + "|" + (string)iAuth, kMenuID);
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
Nice work here. Ties in with #944 and implements LSD functions. Likely to need some changes when we finalize the settings standards with LSD.

tayaphidoux:
your linkset_data() section has more work than is needed there is no reason for the script to have to store this information just call it where its needed, the only part of linkset_data() that is needed is the lock section detach call

Yosty7B3:
> your linkset_data() section has more work than is needed there is no reason for the script to have to store this information just call it where its needed, the only part of linkset_data() that is needed is the lock section detach call

We could remove those but then we have to make sure the linksetdata is set to the right values for all of them.  
Right now their default values are not empty/zero, so if you try to read their linksetdata while it's not set then we'll end up with the wrong default value.

tayaphidoux:
you are right on that i went over the script last night doing just that i am trying to convert the full core set to use LSD the ones that have a default value need to be set by the first script to boot or by the script that would naturally set them, so i came up with the following.

```
FirstRun()
{
    if(llLinksetDataRead("auth_wearer") == "")
    {
        llLinksetDataWrite("auth_wearer",llGetOwner());
    }
    else if(llLinksetDataRead("auth_wearer") != llGetOwner())
    {
        llLinksetDataReset();
    }
    if(!(integer)llLinksetDataRead("global_firstrun"))
    {
        llLinksetDataWrite("global_safeword","RED");
        llLinksetDataWrite("global_prefix",llToLower(llGetSubString(llKey2Name((key)llLinksetDataRead("auth_wearer")),0,1)));
        llLinksetDataWrite("global_channel","1");
        llLinksetDataWrite("global_weareraddon",(string)TRUE);
        llLinksetDataWrite("global_addonlimit",(string)TRUE);
        llLinksetDataWrite("auth_limitrange",(string)TRUE);
    }
    else
    {
        if(llLinksetDataRead("global_safeword") == "")
        {
            llLinksetDataWrite("global_safeword","RED");
        }
        if(llLinksetDataRead("global_prefix") == "")
        {
            llLinksetDataWrite("global_prefix",llToLower(llGetSubString(llKey2Name((key)llLinksetDataRead("auth_wearer")),0,1)));
        }
        if(llLinksetDataRead("global_channel") == "")
        {
            llLinksetDataWrite("global_channel","1");
        }
        if(llLinksetDataRead("global_weareraddon") == "")
        {
            llLinksetDataWrite("global_weareraddon",(string)TRUE);
        }
        if(llLinksetDataRead("global_addonlimit") == "")
        {
            llLinksetDataWrite("global_addonlimit",(string)TRUE);
        }
        if(llLinksetDataRead("auth_limitrange") == "")
        {
            llLinksetDataWrite("auth_limitrange",(string)TRUE);
        }
    }
    if(llLinksetDataRead("global_verbosity") == "")
    {
        llLinksetDataWrite("global_verbosity","1");
    }
}

```

tayaphidoux:
and yes in that i added 2 new values we only need one script to get and check the wearer key this reduces the redundency and if the key is wrong i have it clear linkset which should force a reset of every script if you have it set up. LINKSETDATA_RESET action.

tayaphidoux:
i forgot to add llLinksetDataWrite("global_firstrun",(string)true); under the first run check

Medea-Destiny:
@tayaphidoux I think we should be looking to remove on attach restarts altogether, so we could be doing first run settings in state_entry

Medea-Destiny:
Slightly OT but as it comes up in @tayaphidoux 's comments -- I strongly suggest we no longer use lists for trusted/owner/blacklist auth values. If we save users in the form token: auth_(uuid) value: iAuth we have a faster and more efficient auth process.

Yosty7B3:
instead of adding a firstrun token, we could also just revert back to the default value when a specific token doesn't exist.

        if(!~llListFindList(llLinksetDataListKeys(0,0),[sToken])){ // token/key doesn't exist in the list of keys
            llLinksetDataWrite(sToken, sDefaulVal);
        }

tayaphidoux:
@Yosty7B3 that looks better than what i wrote and is more flexible.

@Medea-Destiny the reset i mentioned is not on worn its on key change and its on LSD Reset, in order to prepare the collar for a new user when its passed from one person to the next, not  reset every script on wear, not that it matters if you put the important stuff in LSD.

Medea-Destiny:
@tayaphidoux If we're clearing LSD then we're effectively doing a "factory reset" and there's no reason not to reset all scripts.

tayaphidoux:
@Medea-Destiny so you would rather some one hand you a collar pre configured the way they want to controll you?

and yes its a factory reset hence why the code below should exist in all linkset blocks, so that the scripts know to repopulate pre defined peramiters, this means the only time you need to reset the scripts are A) when some one clears memory, b) when you pass the collar from one person to the next to  prevent passing maillisious/broken configurations.

```
linkset_data(integer iAction, string sName, string sValue)
{
        if(iAction == LINKSETDATA_RESET)
        {
                llResetScript();
        }
}
```

Medea-Destiny:
@tayaphidoux "@Medea-Destiny so you would rather some one hand you a collar pre configured the way they want to controll you?"

What? Were you thinking we wouldn't clear LSD on a changed owner? I have no idea what you're talking about here.

tayaphidoux:
@Medea-Destiny i am talking about doing just that, and encase a third party script wipes the data.

Medea-Destiny:
A third party script could wipe all settings today just as easily.

PR #944, opened 2023-04-03
Create oc_weld

Yosty7B3:
Replaces the functionality of Weld but in it's on script now (was spread between oc_core and oc_settings).

Using LinksetData instead of prim description for settings, which also allows the Weld button from being dynamically added and removed. 

Most weld dependencies in oc_settings should now be redundant and could be removed.  Already removed (commented out) weld from oc_core in #943

### DIFF
--- a/src/collar/oc_weld
+++ b/src/collar/oc_weld
@@ -1,47 +1,28 @@
-string g_sSubmenu = "Main";
-string g_sButton = "Weld";
+string g_sApp = "Weld";
 
-integer g_iRespring;
+key g_kWelder;
 
-//MESSAGE MAP
-//integer CMD_ZERO = 0;
 integer CMD_OWNER = 500;
 //integer CMD_TRUSTED = 501;
 //integer CMD_GROUP = 502;
-//integer CMD_WEARER = 503;
-integer CMD_EVERYONE = 504;
-//integer CMD_RLV_RELAY = 507;
-//integer CMD_SAFEWORD = 510;
-//integer CMD_RELAY_SAFEWORD = 511;
-//integer CMD_NOACCESS=599;
+integer CMD_WEARER = 503;
+//integer CMD_EVERYONE = 504;
 
 integer NOTIFY = 1002;
 integer NOTIFY_OWNERS=1003;
 //integer REBOOT = -1000;
 
-integer LM_SETTING_SAVE = 2000;//scripts send messages on this channel to have settings saved
-//str must be in form of "token=value"
-//integer LM_SETTING_REQUEST = 2001;//when startup, scripts send requests for settings on this channel
-//integer LM_SETTING_RESPONSE = 2002;//the settings script sends responses on this channel
-//integer LM_SETTING_DELETE = 2003;//delete token from settings
-//integer LM_SETTING_EMPTY = 2004;//sent when a token has no value
+integer LM_SETTING_SAVE = 2000;
+
 
 integer MENUNAME_REQUEST = 3000;
 integer MENUNAME_RESPONSE = 3001;
-integer MENUNAME_REMOVE = 3003;
+//integer MENUNAME_REMOVE = 3003;
 
-//integer RLV_CMD = 6000;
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
Hi Yosty, great stuff, separating out weld into its own script is much needed. We're likely to need some minor changes when we've finalized the changes to settings management and dynamic menus.

As part of the process of moving over to LSD for this (and other settings) we have to take some care with dependencies and redundant code, so this will also rely on changes to OC_settings as you say. We're also going to have to think about the update process here -- I'd anticipate the updater will need to ensure the weld script is added if there's a weld present.

On specifics of implementation: I'd suggest we want this script in the Apps menu, and maintain the unweld code within this script. I think we could have a button in the menu that explains the unweld process too -- "how do I unweld" is one of the most commonly asked support questions. I'm not sure we actually need to use the linkset_data event here at all -- it's something we should avoid if we can for the sake of script efficiency, as as our LSD is easily read globally I'd be inclined to manage the LOCK button behaviour in main menu by checking the intern_weld value locally in the oc_core script.

Yosty7B3:
The LOCK button behavior is still controlled by the core script. and reading the weld flag like you suggest. It's one of the few pieces of weld code I left in core. But thelinkset_data event is for adding/removing the Weld button. I wouldn't know how else to do it tbh.

Yosty7B3:
Come to think of it, if we put Weld in it's own Apps menu, and no longer show the weld button in the main men, then there no longer would be a need for the linkset_data event.  
Catching the event was specifically intended to update the main menu after a respring so that it reflected changes .

Medea-Destiny:
So this is an important and valuable project, and I think we should target it for 8.3.  Weld in its current form causes an awful lot of the support load for OpenCollar.

At the moment this app brings the huge advantage that moving it to an app means we're taking the button a little further away from unwise temptation (we've all seen the "I accidentally welded my collar, I was curious to see what it does and now I can't unlock" complaints), and means we can clear some space in the core and api, which is certainly useful.

Once we've taken the step of providing an app it makes sense to me that the weld app also handles unwelding, and we do without the necessity of an external welder altogether. This can work in exactly the same was as the weld function: the unweld has to be initiated with someone with CMD_OWNER auth, and then the wearer gets a consent popup. This entirely removes the complexity of an external welder.

I have prepared a modification of @Yosty7B3 's OC_weld to do this and remove llLinkSetData dependencies for full compatibility with the 8.3 API which can be found here (as yet untested) https://github.com/OpenCollarTeam/OpenCollar/blob/Medea_destiny_weld_patch/src/Apps/oc_weld . Further work that needs to be done here for the 8.3 inclusion is to remove the weld bits that are no longer necessary in oc_core/api. @Yosty7B3 I assume you've done some work on that already?

I have two extra considerations we could use some input into:

1. Installer. It might make sense to modify the installer so that the oc_weld app is automatically installed into collars that are welded. That means that welded collars can be unwelded by the normal process without modification. Alternatively, we can simply instruct people who wish to unweld to run the updater and make sure they accept the optional oc_weld app. On the downside this is an extra step, on the upside people might actually prefer not having the unweld option right there.
2. We might decide that having an external unweld object is actually a good thing, because the requirement of using an external tool makes the weld process "feel" more real. If so we could simply  scan for any object called "Open Collar Unwelder" within 5 meters and not proceed to the unweld consent stage unless it's present. This way we don't even have to bother with a scripted unwelder.

Yosty7B3:
>  @Yosty7B3 I assume you've done some work on that already?

Yes, I did for the LSD version. It only needed an `if (llLinksetDataRead(... ` to enable the 'welded by' message in the main menu.  For 8.3 that would need to keep more. But all in all it should be pretty simple to make it work with your adjustments. 
I assume we do still want that to show in the main menu when someone is welded.

Medea-Destiny:
@Yosty7B3 Yes, so long as the plugin is there, it should show in the menu. People can always remove the plugin after welding, and add back in if they need to weld, if they wish!

Medea-Destiny:
Discussion on non-technical aspects of this (decisions on how it all works for the user etc) here: #1009

PR #947, opened 2023-04-12
Update oc_rlvsuite.lsl

Medea-Destiny:
Stray restriction sets sit instead of sittp (instructions say it stops you sitting on things unless you're standing right next to them, which is sittp, it currently sets sit). This fix changes the stray restriction values for sittp rather than sit. 

Untested in world so far, someone just brought it up in the group so I put in the fix, so needs checking before approving the pull, but there shouldn't be an issue unless I've typod or done something dumb!


SilkieSabra:
This looks tested inworld, approved and ready to merge.  Any reason there is a Do Not Merge yet label on it still?

Medea-Destiny:
Nope, do not merge was just there pending in world testing!

Issue #953, opened 2023-05-10
oc_badwords does not trigger after transfer


Phuk-SLAM:
What version of OpenCollar are you using?
8.2.3

What behavior did you expect?
Badwords triggers, as expected.

What behavior did you see instead?
Badwords works for the initial owner/creator of the collar. When transferred to another person, the punishment will not trigger, although the next owner can set up the app with trigger words, penance, anim, sound etc, the actual punishment silently fails.

What steps does someone need to take to reproduce the problem?
Add oc_badwords to a collar and transfer it to another user.

SOLUTION FOUND:
oc_badwords does not reset on owner change. When adding

if(iChange & CHANGED_OWNER)  llResetScript(); 

to the changed event, the issue is resolved.

Phuk-SLAM:
modified script attached
[oc_badwords_transfer-fix.txt](https://github.com/OpenCollarTeam/OpenCollar/files/11446789/oc_badwords_transfer-fix.txt)

Medea-Destiny:
Thanks Phuk! Don't worry about the PR, I'll sort that out for you. You'll get the credit though!

PR #954, opened 2023-05-11
Update oc_badwords.lsl

Medea-Destiny:
See #953 

Fix tested by Phuk in world and seems to fix problems, but should be double checked by testing team.

Thanks Phuk!


Pingout (approved):
We tested this and found no problem. Works as advertised.

PR #963, opened 2023-06-16
free up memory in oc_core

Yosty7B3:
A unique identifier key was send with DIALOG which was used to track which script send the Dialog and which menu it was calling. But this same functionality can be achieved by sending that information directly as the identifier key.

The actual logic to keep multiple menu users separate happens in oc_dialog, which creates separate channels for each Dialog call, but the identifier key is not used for this. They identifier key is not used at all in oc_dialog, but only passed unchanged back into DIALOG_RESPONSE. So it seems that anything could work as a key without changing functionality.

With this change the identifier is the name of the menu + the name of the script: sMenu+llGetScriptName(). From what I can tell, this should be sufficient to keep it working as before without having to store the list g_lMenuIDs.

### DIFF
--- a/src/collar/oc_core.lsl
+++ b/src/collar/oc_core.lsl
@@ -115,12 +115,7 @@ key g_kWeldBy;
 list g_lMainMenu=["Apps", "Access", "Settings", "Help/About"];
 
 Dialog(key kID, string sPrompt, list lChoices, list lUtilityButtons, integer iPage, integer iAuth, string sName) {
-    key kMenuID = llGenerateKey();
-    llMessageLinked(LINK_SET, DIALOG, (string)kID + "|" + sPrompt + "|" + (string)iPage + "|" + llDumpList2String(lChoices, "`") + "|" + llDumpList2String(lUtilityButtons, "`") + "|" + (string)iAuth, kMenuID);
-
-    integer iIndex = llListFindList(g_lMenuIDs, [kID]);
-    if (~iIndex) g_lMenuIDs = llListReplaceList(g_lMenuIDs, [kID, kMenuID, sName], iIndex, iIndex + g_iMenuStride - 1);
-    else g_lMenuIDs += [kID, kMenuID, sName];
+    llMessageLinked(LINK_SET, DIALOG, (string)kID + "|" + sPrompt + "|0|" + llDumpList2String(lChoices, "`") + "|" + llDumpList2String(lUtilityButtons, "`") + "|" + (string)iAuth, sName+"~"+llGetScriptName());
 }
 integer g_iHide=FALSE;
 integer g_iAllowHide=TRUE;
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
Won't this cause conflicts if two users are attempting to use the same menu?

Yosty7B3:
I've gone through the different steps and it seems to me there is no reason why there would be a conflict with 2 users using the same menu.

The dialog messages contain the key of who called the menu, so the scripts already know who is who.  
oc_dialog also keeps a (much more elaborate) list to keep track of the the dialog calls, and the identifier for that is the channel number.  So different users are separated through the channel, which links to all the data connected to that dialog.

The identifier key is only stored in it to pass on later. The only time it's used in oc_dialog is to send as key with DIALOG_RESPONSE and DIALOG_TIMEOUT . So it's basically a clean pass through.

tayaphidoux:
should be fine if the code is done properly, i found out that the ao menu code origionaly was not and had to fix a flaw in how it cleared information, since it caused a memory build up, as you navigated the menu.

NikkiLacrima:
The last argument in llMessageLinked is a key, strings can be supplied when keys are expected but in lsl specification they should be 36 hexadecimal charcters or dashes for valide keys. Sending a string "menu~oc_core", a unique string but not a proper UUID, where a key is expected is not guaranteed to work with future updates of the SL compiler and runtimes.  

I suggest not accepting this PR.

tayaphidoux:
@NikkiLacrima  i have seen this done countless times in countless other scripts over the years, what makes you sure they are planning to change this portion of lsl code any time soon?

NikkiLacrima:
Of course I am not sure what they plan, and tbh I didnt read that note, just checked the sinature of llLinkMessage, its rather silly of LL to specify that field as a key if their system never treats it as such. just any old string. In that case removing the kMenuIDs is a good thing and I agree. :)

tayaphidoux:
@NikkiLacrima  the thing is jsut passing it to a key as you would be expecting a key results in NULL_KEY so you just have to check if its a NULL_KEY if its nto use it like a key other wise its a string or blank.

NikkiLacrima:
That would be type conversion and would need a type conversion specification: (integer)sSomeString, otherwise the compiler in my experience complains about wrong type of arguments.

tayaphidoux:
@NikkiLacrima  its what happens when you pass the string in key to a key it type converts it automaticaly.

NikkiLacrima:
No, then all non UUID strings would be converted to NULL_KEY I think, but must check that in code

tayaphidoux:
@NikkiLacrima the difference is if(kID != NULL_KEY){ kID is a key) else if(kID != ""){kID is a string}

NikkiLacrima:
Agreed, but testing for null values is not issue here, anyway, I think the propsed method is workable and could use some more testing to make sure.

Medea-Destiny:
@NikkiLacrima I'm pretty confident that LL will never change key variables being a specific instance of a string such that it can be used this way. It would break vast amounts of content, and there would be riots in the streets!

NikkiLacrima:
Yeah I have alreay accepted that, since their link_message event handler documentation says so. Keys and strings are mostly interchangable, well keys are strings with different operators so care must be exercised, so strings can carry key data and keys can carry string data until you try to use operators like == and !=

NikkiLacrima (approved):
Tested inworld, no issues due to this code, approx 600 bytes memory less is good.

Issue #965, opened 2023-07-23
Leash particles persist forever in some cases due to an incorrect distance calculation in oc_particle that results in the LockGuard 'unlink' command being ignored


cudatox:
What version of OpenCollar are you using?
8.2.3

What behavior did you expect?
The collar should respond correctly to the 'unlink' command by removing the particles connecting the collar to the target

What behavior did you see instead?
The particles continue to persist forever

What steps does someone need to take to reproduce the problem?

This was tested using the LockGuard plugin for AVSitter 2.2. Simply sit on any piece of furniture that is using this plugin and is located far from the region origin. When you stand up, the collar will not unleash itself. I have also noted this issue on some furniture not using AVSitter.

The issue seems to lie in the manner in which the distance calculation is preformed in oc_particle (~line 756). On some furniture, LockGuard does not send the chain target (parsed into kLGTarget) when sending the unlink command to the collar. This results in kLGTarget being assigned NULL_KEY and when it is subsequently used in the distance calculation, it results in an incorrect distance value that may exceed the limit (g_iLeashLength). This usually results in the unlink command being ignored.

I was able to correct this issue in my experimentation by bypassing the leash length check for the unlink command, but I'm not sure what the other potential consequences of doing this would be:

````
               if(llVecDist(llGetPos(), (vector)llList2String(llGetObjectDetails(kLGTarget, [OBJECT_POS]),0)) > g_iLeashLength){
                    //Leash length is irrelevant for the unlink command
                    if (sLGCMD != "unlink"){
                        return;
                    }
                }
````

Pingout:
This may also be a solution to https://github.com/OpenCollarTeam/OpenCollar/issues/650
#650 
I will test it and report back

NikkiLacrima:
Handled with PR #974

Issue #650, opened 2021-08-25
cuff points collar release code.


tayaphidoux:
What version of OpenCollar are you using? 
V8.1.0000

What behavior did you expect?
when released from something that connects a chain to the back left and right leash points for it to remove said chain

What behavior did you see instead?
the chain is still there even after the item was drezed and a relog

What steps does someone need to take to reproduce the problem?
ping's leash point tester, and maybe any other rlv item that takes those points 
on the collar.

Pingout:
Superceeded by https://github.com/OpenCollarTeam/OpenCollar/pull/966

PR #966, opened 2023-07-23
Update oc_particle.lsl

Pingout:
Potentially a fix to #965 and #650


Trinkitz:
I tested this today using a vendor collar; I also did the test using the free OC cuffs. The chains still persisted for me on both the collar and cuffs. It may be useful to point out that even though the leash particle shows, I am not leashed; the collar menu says I can move freely. I can clear all of this by leashing myself to something and then unleashing myself. Hope this helps.

Pingout:
Superceeded by PR #974

PR #974, opened 2023-08-04
Update LockGuard handling in oc_particle

NikkiLacrima:


Pingout (approved):
From my testing on multiple devices, from a number of different vendors, this appears to have resolved the "empty sprite" chains we have been experiencing, along with ending the chains upon standing from these devices

Issue #935, opened 2023-02-08
group auth not being cleared by runaway?


Medea-Destiny:
Reported to me in world by Clody (gynger81). Have not yet investigated, putting this here to remind me to. Anyone else observed this behaviour?

Pingout:
Trinkitz and I tested this a couple of times, and yes..... Group Access is not cleared by Runaway command.  However the runaway command makes the wearer own themselves, and they can simple uncheck the Group Access box themselves.

SilkieSabra:
Minor issue but I think for consistency it should be corrected.  All access should be cleared by runaway.

Issue #399, opened 2020-08-27
Integrated V 7.5 Relay allows global safeword in Helpless mode


SilkieSabra:
Tested the integrated relay with Auto and Helpless Set.

Helpless should both prevent the REFUSE button from showing up in the relay menu, and also disable the global collar safeword from releasing the relay.

It did stop REFUSE but the global safeword still ended the trap.

zontreck:
Not a bug
Aug 27, 2020 12:54:43 PM Silkie Sabra <notifications@github.com>:

zontreck:
This is a feature. Collar safeword should not be limited unless safeword is disabled globally

PR #975, opened 2023-08-07
Update oc_api.lsl

NikkiLacrima:
Clear group auth on runaway, this should solve issue #935


Pingout (approved):
This change works as expected, clearing Group Access checkbox on Runaway.
Only a single line has changed in the script (Line #399) so it looks good as far as Code Review"

PR #977, opened 2023-08-11
Decouple bell sounds with bell visibility

AdmiralTails:
"Bell is ringing and hidden." is currently a lie because of this. Also they're separate options and should operate separately.


Pingout:
I'm not clear on what you are trying to do. Are you trying to make the bell ring when the bell is not visible? I don't understand why we would want to do that. Can you provide an explanation of what you hope to fix? Thanks

SilkieSabra:
Can you please transfer this pull to the [8.3 features branch](https://github.com/OpenCollarTeam/OpenCollar/tree/8.3_Features-branch) ? We are currently using the master branch as our stable repository, the working version accepting changes to the code is the next version branch, which will be 8.3.

AdmiralTails:
Will do.

PR #978, opened 2023-08-11
 Decouple bell sounds with bell visibility

AdmiralTails:
"Bell is ringing and hidden." (the text for bell enabled but hidden) is currently a lie because of this, the bell will not ring on movement if the bell is not also toggled on (which may not be set up, or even able to be properly set up on some collar meshes).

Also they're separate options and should operate separately.


Pingout:
My comment is similar to the comment I made for PR #977.  Are you trying to make the bell ring when the bell is not visible? I don't understand why we would want to do that.
The text you mention is the Menu Promp in the dialog of the Bell Menu. I think the problem is incorrect logic in setting up the Bell Menu Prompt. When the bell is hidden, the sound should be off and the Menu Prompt should reflect that state.

AdmiralTails:
A few reasons why someone would want the bell to ring when it is not visible:
1. The bell's mesh on their collar is not set up for the Bell app's visibility toggle, and potentially cannot be because the unscripted or other scripted collar they have added OC scripts to does not have the bell as a separate object. This prevents them from toggling visibility. 
2. Their collar does not have a bell on it, but they have attached a separate object that does, and wish to have the already existing bell app in OC handle the sound.
3. The bell would realistically be underneath something else they are wearing (such as a scarf or shawl), but clips with said object, so they have hidden the bell's mesh while wearing it.

As an additional note, this behavior currently isn't even consistent. When the bell is hidden but on, it will make sounds on collisions, but not when walking/running.

tayaphidoux:
@AdmiralTails from what you mention there would be no bell prim itself listed for the rendering of visible or invisible no objectdetail with Bell there fore you could check for an object with that if no object has Bell in its description then you could flag a toggle 
changing the if(!g_iBellShow)return; to if(!g_iBellShow && !g_iNoBell) return;

AdmiralTails:
@tayaphidoux That could work if it were desired to keep the current behavior when a bell is defined. I do not believe that that behavior should be maintained, however. The bell app's menu presents the sounds and visibility as fully separate toggles. Currently, one configuration of those toggles, in which bell sounds are enabled but the bell is hidden. does not behave in the manner it presents itself, with movement-triggered sounds still not playing, while collision-triggered ones still do.

As the menu currently presents its settings, the fact this check exists at all is a bug.

Medea-Destiny (approved):
This code will as intended stop the bell from ringing on movement if the bell is on but hidden.  Decision to merge should be on desirability of function.

PR #980, opened 2023-08-13
oc_dialog - Invalidate menus for removed owners

NikkiLacrima:
This patch checks for changes in the owner list and if an owner is removed calls ClearUser() to invalidate any open dialogs for that user.


Medea-Destiny (approved):
Code looks good to me. @NikkiLacrima can you add a credit note to the top of the script with your name, date, short description of the change and reference to the issue fixed please?

PR #981, opened 2023-08-13
Combine all oc_core menu functions

Yosty7B3:
Moving all the code from the separate menu functions into the dialog() function to free more memory in addition to #963

Should function exactly as before just taking less memory.


Medea-Destiny (approved):
Changes make perfect sense to me. I can't see any problems via eyeball, but we'll need to test in world with particular emphasis on menu respring behavior (i.e. making sure the menu reopens after button presses when it should).

Pingout:
This all tests out very well inworld. Menu resprings work perfectly. I can approve this once @Yosty7B3 makes the documentation change requested by @Medea-Destiny

Pingout (approved):
This is great, thanks!

PR #991, opened 2023-09-11
oc_titler bugfix

Medea-Destiny:
Zero length titles in chat command (i.e [prefix] title) was returning "title" as a title, rather than opening the textbox as intended.


NikkiLacrima (approved):
Looks good, tested with no issues

PR #997, opened 2023-10-17
Adds a addon bridge to hopefully fix #988

### LINKED ISSUE #988: Stand Alone Addon Proposal
THIS IS A PROPOSAL NOT A REQUEST, NOT AN ISSUE.
https://opencollar.cc/docs/Development-Life-Cycle
```
Feature Decisions

Someone’s got to decide what we’ll include in OpenCollar. The founders (Athaliah Opus and Nirea Mercury) have final decision-making power there, but most features don’t start with them. Most often, someone in the community has an idea, they propose it (either in chat or even better, as a pull request), and if there’s consensus that it’s something we want in the collar, it gets added.

That consensus is the hard part. There are a lot of competing concerns at play:

    Is this the kind of feature that we want in OpenCollar?
    Will this feature be used by most users, or just a small subset?
    How much does this new feature complicate OpenCollar? Is the additional functionality worth that additional complexity?
    Is this going to be a surprising change to how existing collar users expect things to work?
    Who’s going to do the work to add this?

If you want to propose a new feature for OpenCollar, there are a couple things you can do to improve your odds of success:

    Communicate clearly, in durable written form. If it’s a small feature, then a paragraph or two in a Github issue might be all you need. If it’s a big feature, or especially if it’s a complicated architectural change, then you might need to create a separate document or diagrams to make it easy to understand.
    Seek feedback. Discuss your proposal with the project leaders and be prepared with answers to the questions above. The OpenCollar Discord server is the best place for this. Consider gathering feedback from the larger community, such as through group chats or a poll.
```

(edit note: added this because i think some people are confused of why this is in issue.)

i Closed the previous pr because of https://github.com/OpenCollarTeam/OpenCollar/pull/960#issuecomment-1704392375 this will clear up the pr and allow me to discuss this template with others as i work to incorporate what users might be interested in.

```
define "stand alone":
Intended, designed, or able to be used or to function alone or separately : not connected to or requiring connection to something else in order to be used or to function.
https://www.merriam-webster.com/dictionary/stand-alone
```
Rssentially the add-on would function without the collar connection but could be connected to the collar, like the pending AO.

```
define "template":
1.  A gauge, pattern, or mold (such as a thin plate or board) used as a guide to the form of a piece being made
2.  Something that establishes or serves as a pattern
https://www.merriam-webster.com/dictionary/template
```

the current version in that closed pr is functional but there has been an expressed desire for it to 1:1 compatible with oc_plugin template so that you could drop in oc_plugins and core files to help create templates, it will take time and more work than i have put in at the moment, the stand alone add-on template started out as a template for a stand alone add-on.

The idea at first was to just do lock/unlock which could be dropped in for harnesses and other items to make them lockable, it still serves this function with oc_standalone_addon and oc_addon_menu, but currently has a custom api to create plugins for it
that could load their own menu buttons into the main menu as well as relay information between the collar and said plugin.

The objective is to make something that can allow new creators to drop and use, while giving something to more advanced creators to build from/modify/improve on, i feel that moving to a more 1:1 collar plugin stand alone platform may actually do this so to comply with https://opencollar.cc/docs/Development-Life-Cycle i am closing the pr and moving to issues as mentioned in this guide.

Please feel free to ask me for script updates, or get them from https://github.com/tayaphidoux/OpenCollar-Alpha/tree/8.3_Features-branch -  i will try to keep it up to date with the changes in feature branch as i make changes.

Do interact with me here not the closed pr going forward, the template name may end up changing due to the 1:1 nature desired i am thinking oc_standalone_addon_core and oc_standalone_addon_menu for the main scripts, auth is a plugin for managing users list, as is buttonloader they are examples of the current api as well as a way to test the menu input feature.

zontreck:
This is a testing tool I had created awhile back when i first created the addon system. It sounded like it might resolve the issue.

### DIFF
--- a/src/collar/oc_addons.lsl
+++ b/src/collar/oc_addons.lsl
@@ -423,7 +423,6 @@ state active
     link_message(integer iSender,integer iNum,string sStr,key kID){
 
         if(llGetListLength(g_lAddons)>0){
-            // The below code will create a race condition where an addon might function perfectly, but if an addon that has that filtering, but it is improperly opting in, it will suddenly not work. - Aria
             if(llListFindList(g_lAddonFiltered, [iNum])!=-1){
                 // check if any addons want this link number
                 // filtering list proposed by Caraway Ohmai via Discord
--- a/src/collar/oc_addons.lsl
+++ b/src/collar/oc_addons.lsl
@@ -423,6 +423,7 @@ state active
     link_message(integer iSender,integer iNum,string sStr,key kID){
 
         if(llGetListLength(g_lAddons)>0){
+            // The below code will create a race condition where an addon might function perfectly, but if an addon that has that filtering, but it is improperly opting in, it will suddenly not work. - Aria
             if(llListFindList(g_lAddonFiltered, [iNum])!=-1){
                 // check if any addons want this link number
                 // filtering list proposed by Caraway Ohmai via Discord
--- a/src/spares/oc_addon_bridge.lsl
+++ b/src/spares/oc_addon_bridge.lsl
@@ -41,8 +41,8 @@ integer CMD_EVERYONE        = 504;
 //integer CMD_NOACCESS        = 599;
 
 //integer LM_SETTING_SAVE     = 2000; //scripts send messages on this channel to have settings saved, <string> must be in form of "token=value"
-//integer LM_SETTING_REQUEST  = 2001; //when startup, scripts send requests for settings on this channel
-//integer LM_SETTING_RESPONSE = 2002; //the settings script sends responses on this channel
+integer LM_SETTING_REQUEST  = 2001; //when startup, scripts send requests for settings on this channel
+integer LM_SETTING_RESPONSE = 2002; //the settings script sends responses on this channel
 //integer LM_SETTING_DELETE   = 2003; //delete token from settings
 //integer LM_SETTING_EMPTY    = 2004; //sent when a token has no value
 
@@ -50,6 +50,9 @@ integer DIALOG          = -9000;
 integer DIALOG_RESPONSE = -9001;
 integer DIALOG_TIMEOUT  = -9002;
 integer REBOOT = -1000;
+
+integer ALIVE = -55;
+integer READY = -56;
 integer STARTUP = -57;
 
 
--- a/src/spares/oc_addon_bridge.lsl
+++ b/src/spares/oc_addon_bridge.lsl
@@ -1,0 +1,212 @@
+// This file is part of OpenCollar.
+// Copyright (c) 2009 - 2016 Cleo Collins, Nandana Singh, Satomi Ahn,   
+// Joy Stipe, Wendy Starfall, Medea Destiny, littlemousy,         
+// Romka Swallowtail, Garvin Twine et al.  
+// Licensed under the GPLv2.  See LICENSE for full details. 
+
+// * This file is GPL because it is considered feature complete, and should work out of the box to integrate plugins as an addon.
+// In discussions with Nirea originally, the template was determined to be public domain, because it was not feature complete, and required edits by the end user to create a product.
+
+/*
+-Authors Attribution-
+Aria (Tashia Redrose) - (August 2020 - October 2023)
+Lysea - (December 2020)
+Taya'Phidoux' (taya.maruti) - (july 2021)
+*/
+
+integer API_CHANNEL = 0x60b97b5e;
+
+//list g_lCollars;
+string g_sAddon = "-notset-bridge-";
+
+//integer CMD_ZERO            = 0;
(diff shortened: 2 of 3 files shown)


zontreck:
This statement hereby marks my agreement to the CLA for this contribution.

tayaphidoux:
first you are doing the very thing as to why i closed my pr and moved it to issues. 

@SilkieSabra if you allow this and not my system to proposed this way you are picking and choosing and the rules are for not.

https://opencollar.cc/docs/Development-Life-Cycle
```
If you want to propose a new feature for OpenCollar, there are a couple things you can do to improve your odds of success:

    Communicate clearly, in durable written form. If it’s a small feature, then a paragraph or two in a Github issue might be all you need. If it’s a big feature, or especially if it’s a complicated architectural change, then you might need to create a separate document or diagrams to make it easy to understand.
```

second while this does bridge oc_plugin to oc_addon it does not work in stand alone and does not replace what i am working on, it also would allow potential conflict between the oc_plugin in the collar and the same oc_plugin in the addon, duplicating response which is why highlander exists to prevent two collars existing at once, this would work fine if you want to move an oc_plugin script from the collar into an addon for example fashion collars with hidden core, and would be a lighter solution than mine, it would not work for a bridle system, where the leash is supposed to be separate from the collar, or even visibility, and textures etc.

there is no issue with this system and my system existing at the same time.

zontreck:
There may have been a misunderstanding as you stated you wanted it to effectively do this, but what wasn't clear was you meant self-contained standalone. These two systems are different. I can resubmit the PR if that would make you feel better, without a Fix commit, and instead a reference to the issue. I'll wait to hear what roan would prefer. I only intended to be helpful with this, as it was something I'd already made. I did not intend to upset you.

tayaphidoux:
@zontreck  i like your script its fine what i don't like is if you are allowed to just submit it like this where i got told i had to follow the guide for something like this, its not you its the system if one has to follow it i expect all have to follow it other wise there is no point of having it.

it is simpler for cases where some one wants an addon that stays connected all the time and requires the collar to be present and it makes it smaller, and simpler just runs into the issues i mentioned that created the reason for highlander which i put in effort not to duplicate.

tayaphidoux:
@zontreck these change look like they solve the issue of the oc_plugin not starting up when added after collar startup, still no way to filter out and prevent collision in the event of two oc_plugin scripts of the same name and type existing in apps though and there is potential of the two plugins sharing the same input for example prefixleash i recommend binding command calls to the add-on name so it turns into prefixaddon leash

zontreck:
I've been intentionally not responding, but your latest comment shows how very little you know about the collar system. This is an addon, meaning it is not installed within the collar. Plugins added after would be caught by the changed event. The changes I made weren't to try to address your "concerns" about collision. They were to add missing code from the original, which existed before the addon template was created and was significantly more messy. 

The change made that you appear to be referring to simulates a collar startup within the addon so that plugins which depend upon hearing the STARTUP signal will begin to start.

While binding to command calls is good in theory, some scripts, such as leash, do not reside in the Apps menu, and would therefore not even function properly without some changes to the bridge. Changes would be required in the leash script to allow multiple copies of itself to co-exist. It would conflict. The moment that LM_SETTING_SAVE is called, there would be duplicate leashes from all oc_leash scripts, in the collar, and addons using this bridge. 

The use-case here is scripts that do not exist already within the collar. Again, someone would need to make changes to leash to prevent that possible race condition. It may only trigger after a reboot, but it would still be inevitable, and would result in unpredictable behavior.

tayaphidoux:
@zontreck " While binding to command calls is good in theory, some scripts, such as leash, do not reside in the Apps menu, and would therefore not even function properly without some changes to the bridge. Changes would be required in the leash script to allow multiple copies of itself to co-exist. It would conflict. The moment that LM_SETTING_SAVE is called, there would be duplicate leashes from all oc_leash scripts, in the collar, and addons using this bridge. " 

this is exactly why my system functions as a stand alone to allow creators to use things like leash for things like bridles and fashion collars, i filter out those calls so they are localized to the add-on.

"The change made that you appear to be referring to simulates a collar startup within the addon so that plugins which depend upon hearing the STARTUP signal will begin to start."

precisely ping noticed that the script didn't do any thing when i mentioned this pr and they tried your scripts, weather or not they talked to you about it i don't know.

"The use-case here is scripts that do not exist already within the collar."
this relies on the person using this script understanding that which is my point, the other use case for my system is to lower the bar of entry for developers who are just starting out or in general don't have a lot of scripting skills to do most of the work, allowing them to focus on one part of the system to start with.

zontreck:
This is getting off topic and github is not the place for grievances, if you want to speak, message me in Second Life. I am leaving it at that.

Issue #1003, opened 2023-11-04
Spy app not clearing settings after Runaway, and still transmits location to former owner


Trinkitz:
What version of OpenCollar are you using?
8.2.3
What behavior did you expect?
Spy settings to clear after running away from an owner.
What behavior did you see instead?
Spy settings persisted after runaway, and the former owner could still see the wearer's location, though the former owner retained no other ownership abilities that I could find.
What steps does someone need to take to reproduce the problem?
Install Spy, add an owner and runaway, then check to see if Spy settings are still in place.

PR #1004, opened 2023-11-05
Nikki lacrima patch oc spy

NikkiLacrima:
Fixing the issues #1003 oc_spy does not clear spy reports after runaway or last owner removed.


SilkieSabra:
Needs approving review !

Trinkitz:
Tested this patch as both an Owner/wearer with two additional owners and as a non-Owner/wearer with three owners. Spy settings cleared following Runaway, and there was no further tracking of the wearer. Seems a successful fix. Thanks!

Medea-Destiny (approved):
Code looks fine to me. Sensible quick fix,  for 9,x with the move towards linkset data we'll want to handle runaway centrally anyhow so this looks solid for now.

PR #1005, opened 2023-11-06
Replace "return R()" with "return" in R() function.

neil2ball:
2023-11-06 00:29:39,250 INFO  [YEngine]: - <229.76128, 186.58499, 21.009304> OpenCollar - 8.2.3:oc_folders
2023-11-06 00:29:39,251 INFO  [YEngine]: - (157,58) Error: function returns void, no value allowed
2023-11-06 00:29:39,251 INFO  [YEngine]: - (158,59) Error: function returns void, no value allowed
2023-11-06 00:29:39,252 INFO  [YEngine]: - (159,56) Error: function returns void, no value allowed
2023-11-06 00:29:39,252 INFO  [YEngine]: - (160,57) Error: function returns void, no value allowed
2023-11-06 00:29:39,253 INFO  [YEngine]: - (161,47) Error: function returns void, no value allowed
2023-11-06 00:29:39,253 INFO  [YEngine]: - (209,61) Error: function returns void, no value allowed
2023-11-06 00:29:39,254 INFO  [YEngine]: - (210,62) Error: function returns void, no value allowed
2023-11-06 00:29:39,255 INFO  [YEngine]: - (211,59) Error: function returns void, no value allowed
2023-11-06 00:29:39,256 INFO  [YEngine]: - (212,60) Error: function returns void, no value allowed

LSL is not my area of expertise, but I wouldn't think that you would try to use R() as a recursive function. OpenSim's YEngine does not like this as it views R() as a void.


NikkiLacrima:
The return R() is nit in the R() function, it is in the following Browser(key kID, integer iAuth, string sPath) function.

Original code is correct, and it is not a recursive call.

neil2ball:
Sorry, it was getting late when I was working on this. This must be some kind of error in the YEngine then. I'll forward this to OpenSim.

Pingout:
Is this script change for OpenSIM only? And if so, we can not point it to the Master repo.
Should we start a repository for exclusively OpenSIM script?

NikkiLacrima:
The patch does not change any fuctionality for LSL open collar, and the current `return R();`  construct is formally not correct even if it works for LSL.  So I think the PR in the form of the second commit can be approved, after testing.

NikkiLacrima:
Please use 4 spaces and not tabs, one never knows how many space a tab will be turned into.

neil2ball:
> Please use 4 spaces and not tabs, one never knows how many space a tab will be turned into. Especially when moving around between SL script editor, GitHub webb and other script editors for comparison checks.

I'm not quite sure why it changed the spacing. I just used LSL Editor, but I have a feeling that isn't what people use these days. I'll make another edit to the file to include four spaces instead.

Medea-Destiny:
Agree with @NikkiLacrima that the ```return R();``` though it functions is not good practice and can simply be replaced by ```{R(); return;}```. While not necessary for SL let's do it. 

However the formatting here is problematic with a lot of change to spacing that misaligns the top comments badly once inserted into an SL script, and creates a +573/-573 line change that may cause confusion in future. Can we have an edit of the file as it currently exists in the OpenCollarTeam:8.3_Features-branch [https://github.com/OpenCollarTeam/OpenCollar/blob/8.3_Features-branch/src/collar/oc_folders.lsl](url) with just the return R(); statements replaced?

Medea-Destiny:
Looks good, thanks @neil2ball

NikkiLacrima:
Tested in world, all seems good. Approved

neil2ball:
Sorry if this was a mistake. I'm a button presser.

NikkiLacrima:
I dont know how this can affect the OpenCollar/8.3_Features-branch, one possible way to handle is to reture this PR and I create a new clean branch and PR with the agreed changes, this one has become a jumble of commits that are then overwriting each others.

PR #1006, opened 2023-11-11
Update oc_outfits.lsl

Medea-Destiny:
core lock check needed further fixing, kID rather than kAV was corrected for the actual auth pass, but not for the notify message for no access


NikkiLacrima (approved):
Tested in world, compiles cleanly, sends "no access" notification when trusted tries to lock core. Code inspection ok. Approved.

PR #1007, opened 2023-11-13
Replace "return R()" with "return" in R() function part deux

neil2ball:


NikkiLacrima (approved):
Looks good, approved for merging.

Issue #1001, opened 2023-11-03
runaway problems


NikkiLacrima:
What version of OpenCollar are you using? 8.2.3

What behavior did you expect? Runaway to be executed only when enabled and confirmed by wearer

What behavior did you see instead? Using chat "prefix runaway" any owner and/or wearer can unleash and clear the owners, trusted and blocked lists even when disabled and not confirmed by wearer.

The root cause of this is that the  `llMessageLinked(LINK_SET, CMD_OWNER, "runaway", g_kWearer)` message is used both to try and intialize a runaway and then used, from oc_api, to signal that the runaway is enabled and confirmed by the wearer.
In oc_api the g_iRunawayMode variable is used to distinguish between these two cases, but that is not visible outside oc_api.

I propose that we use `llMessageLinked(LINK_SET, CMD_OWNER, "runaway_confirmed", g_kWearer)` for the latter case. 
The changes for this should be quite minor. And also make the g_iRunawayMode handling unnecessary.

Consequently other scripts should listen not for “runaway” but for “runaway_confirmed”.

The uses of the "runaway" message seems to be in the following places (needs another careful checking of the codebase:

oc_core:220
	Deletes the settings auth_owner, auth_trust and auth_block. This can be removed since its already done in oc_api, and it is a bug that allows any owner to effect auth_ removal without runaway enabled or confirmed

oc_relay : 139
	Dead code that is never run due to the check that text commands start with “menu” or g_sSubMenu

oc_camera: 333

oc_themes, oc_anim, oc_relay, oc_titler
	Boilerplate code that resets owner, trust and blocked lists, but these lists are never set or used so can be removed

oc_leash:491
	This is an actual bug that releases the leash even if runaway is not confirmed

unwelder and unwelder hud
	Simply returns

oc_OwnerOnlineCheck:84
	Simply returns, this code lacks handling of owner removal, se #634

Medea-Destiny:
Hey @NikkiLacrima, checking in on this... runaway_confirmed seems like a solid idea, are you working on this currently?

NikkiLacrima:
Yes I am working on it and current plan is :

Move the all of actual runaway execution, that is the clearing of auth lists, down to the RunAwayMenu "Yes" handler where there already is code that sends runaway link messages. This makes the g_iRunawayMode superflous.

Rearrange the RunAway menu a bit to make permission handling for confirmation Yea/No and enable/disable clear.

[...]

PR #1010, opened 2023-11-19
Nikki lacrima patch runaway

NikkiLacrima:


NikkiLacrima:
Updated runaway handling as discussed in issue #1001 
Also removed g_lMenuIDs in these files see #963

Pingout (approved):
Received (4) script from Nikki
    oc_api
    oc_core
    oc_leash
    oc_relay

After installation, the [prefix] runaway command will bring up a menu for the Wearer
 "Are you sure you want to runaway from all owners?"
and offers a Yes, No, and Disable option

For the Owner, the [prefix] runaway command will bring up a menu offering the
option to Disable 

When the Owner selects the Disable button, the collar Wearer gets an Accept/Deny
option. Once accepted by the Wearer, the setting "auth=runaway~0" is in place and
a further [prefix] runaway from the Wearer results in a notice to the wearer:
 "Access Denied to runaway, or the runaway settings menu"
If the collar owner uses the [prefix] runaway command again, they get a menu with 
the option to Enable runaway

Conclusions: The fix works as expected. Thank you.

Medea-Destiny:
Some preliminary notes:

oc_api line 725, notify owners is sent after owner list is deleted. This line should go above the LM_SETTING_DELETE lines, and it may not be a bad idea to drop in an llSleep(0.5); in case of message timing funkiness.

oc_leash line 488 triggers Unleash function on runaway_confirmed. This function provides notification of who's unleashing what, then triggers the DoUnleash function.  However we also have CMD_SAFEWORD trigger DoUnleash on line 757. I don't think we need the specific unleashing text when a runaway is performed, and the runaway triggers CMD_SAFEWORD, so I don't think we need the runaway_confirmed handling in the UserCommand functioning at 488, and that saves the double call to DoUnleash.

Credits at the top of oc_leash have no detail of changes, and I think all 3 scripts could use a mention of "implemented Yosty's menu streamlining".

Will take a deeper dive soon. Thanks so much for handling this, @NikkiLacrima !

NikkiLacrima:
Medeas comments fixed. Time to merge ?

Pingout (approved):
As a double check, I brought these three updated scripts inworld, and they all tested the same as the last time.

Issue #1008, opened 2023-11-14
Unleashing strict leash removes TPLURE exception for owners


MinesweeperM:
What version of OpenCollar are you using? 8.2.3

What behavior did you expect? Owners can sent TP offers if exceptions are active, and ACCEPTTP restriction is applied.

What behavior did you see instead? Since TPLURE exception removing, during unleashing, TP offers from owner fails.

What steps does someone need to take to reproduce the problem?
1. Activate "Strict" leash option.
2. Activate STRAY restriction
3. Leash
4. Unleash
5. Go to another sim and send TP offer (with owner rights)

Medea-Destiny:
This happens because the TPLURE exception is sent to RLVsys, and removed from RLVsys, while exceptions are set in RLVextension. RLVsys manages sources, which is intended to remove restrictions when a source is released only if there is no other source applying the restriction. In the case of an exception it sees no other source and therefore releases the exception,  which is handled elsewhere.

The simple fix for this would be to have the leash script request a re-triggering of user exceptions when a user unleashes.

MinesweeperM:
Also, this exception removing, when leash holder leaves region, while leashed, so it seems, now it useless even to add it, when grab leash.

Medea-Destiny:
Fixed in 8.,3

PR #1013, opened 2023-11-28
Fix strict leash removing exception

Medea-Destiny:
Issue #1008 

oc_leash: Added EXC_REFRESH call after releasing strict leash to ensure that exceptions that should be in place get restored.

oc_rlvextension: added EXC_REFRESH Link message handling to trigger refresh of exceptions, for above. Also folded bool() function into Checkbox() function and strReplace into MuffleText() functions to free up memory.


Pingout (approved):
Tested by Omnia and Ping. These changes work as expected.
We tested the strictleash with various restrictions set and the script consistently reset the correct restrictions when strictleashed collar wearer was unleashed.
Thank you!

Medea-Destiny:
Conflicts fixed.

Issue #1012, opened 2023-11-20
Height adjust for no pose


Medea-Destiny:
The height offset in oc_anim to allow a stored offset per pose will store an offset if the arrow buttons are used when no pose is active. This can cause confusion. It's possible there are a small number of people who have discovered this and use it to set an offset when they are not being animated, I've seen a number of people who have had mystery offsets they didn't want because of this.

We should fix it so offsets can't be stored for no pose.

PR #1014, opened 2023-11-28
Update oc_anim.lsl

Medea-Destiny:
Added StandOffset checkbox button to Anim menu. This toggles use of hover hieght offset when no pose is selected (i.e. for stand animation). This has been a frequent problem in support, with people accidentally setting a stand offset. However as some people may use this it seemed better to keep the capability but have it switchable and default to off. Issue #1012

### DIFF
--- a/src/collar/oc_anim.lsl
+++ b/src/collar/oc_anim.lsl
@@ -21,6 +21,13 @@ K9K8E
 Tayaphidoux
     Jun 2022    - Restore AO pause functionality
 
+Medea (Medea Destiny)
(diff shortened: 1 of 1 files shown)


Pingout (approved):
Tested by Omnia and Ping. These changes work as expected. The 'standing' height adjustment appears to affect the Hover Hight slider. The individual pose hights are not adversly affected. The 'StandOffset' height goes back to normal when unchecked, and returns to the offset when 'StandOffset' is checked again
Thank you!

PR #1015, opened 2023-11-28
oc_rlvsuite fix chat commands for restrictions

Medea-Destiny:
Chat command for applying restrictions had problems -- if the restriction itself is given rather than the button name, it would break. Added sanity check to ApplyCommand function, index of g_lRLVList is now checked to see if it gets an RLVCMD and if so converts to index of ButtonText, and returns if hit is CategoryIndex. Ensure the full command is fed to ApplyCommand, not just the first word -- ie (prefix) restriction add "TP Location" shouldn't attempt to add just "TP". NOTE: This fix will also allow presets with a space in the name to be accessed properly. 

Changed filtering in UserCommand so that capitalization does not have to match (i.e it  will accept "Preset" as well as "preset". Now also passes rlv as a synonym for setting  restrictions which was coded but rejected before hit, but filters to only pass to   ApplyCommand if there is a command following that is not "on" or "off" to avoid collision  with RLV menu

 Removed double menu for (prefix) restrictions

### DIFF
--- a/src/collar/oc_rlvsuite.lsl
+++ b/src/collar/oc_rlvsuite.lsl
@@ -2,12 +2,14 @@
 //  Copyright (c) 2018 - 2019 Tashia Redrose, Silkie Sabra, lillith xue                            
 // Licensed under the GPLv2.  See LICENSE for full details. 
 medea (medea destiny)
-    June 2021   -   Fix for issue #492 / issue #432, command to update camera settings (blur amount, maxcamera and mincamera)
-                    when numeric value changed via RLV/RLV settings/Camera used insufficent auth level and incorrect 
-                    bitmask values to alter restriction while active even though the setting was altered.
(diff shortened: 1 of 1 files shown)


Pingout (approved):
This tested as expected. RLV Restrictions now work properly with chat commands.
One note though, capitalization in using the Button Name is required. For example '[prefix] restriction add Blur View' will work but '[prefix] restriction add blur view' will not work.
'[prefix] restriction add setdebug_renderresolutiondivisor' works as expected.
The [prefix] restrictions chat command resulted in only one Restrictions Menu

Medea-Destiny:
Added fallback mode for case insensitivity to solve capitalization requirements.

Pingout:
> Added fallback mode for case insensitivity to solve capitalization requirements.

Medea, can you check this please? I'm still having a problem with case sensitive, specifically '[prefix] restriction add blur view' will not work still

Issue #906, opened 2022-11-29
message when login in: you are owned by.... is missing


JuliaC15:
What version of OpenCollar are you using?

What behavior did you expect?

What behavior did you see instead?

What steps does someone need to take to reproduce the problem?

SilkieSabra:
I believe this user meant to report that when starting up, there used to be a message saying your owner's name in chat.  It went missing at some point in V8.   Who would like it back?

Pyhrn:
+1 from me for having it back

Trinkitz:
The message can be restored by running an update, even if that update is the same version as the one on the collar. The message will vanish again the next time the collar restarts, though.

Pyhrn:
I did some digging in the scripts. The message is handled by oc_api (in line 623) and the changelog for 8.2.3 about oc_api says  `Changes: Only reset script when needed, not on each reboot`.  
I suspect that is the reason why the messages appears only after a reboot/update, not on every restart.

Medea-Destiny:
Fixed in 8.3

Issue #897, opened 2022-11-07
The "#" prefix wildcard is not working in 8.2.3 


Pingout:
What version of OpenCollar are you using? Collar 8.2.3

What behavior did you expect? There are two global prefixes. * issues commands to everyone in range with the required level of access, and # issues commands to everyone in range with the required level of access except yourself. (Cut and Paste from the User Guide)

What behavior did you see instead? The # symbol is not working, and does nothing. The * symbol works as advertised.

What steps does someone need to take to reproduce the problem? When close to someone you have collar access to, try the # wildcard in a chat command.

Medea-Destiny:
Looks like another thing that got lost in the 8.x API transition.

Medea-Destiny:
fix in #1016

PR #1016, opened 2023-11-28
oc_api restore interface channel commands

Medea-Destiny:
Fix restores and expands arbitrary command functionality to interface channel. Also restores "owned by:" messages on all start ups, not just reboot, improves auth efficiency of chat commands, and restores "#" prefix handling

Details:
* Restored # prefix functionality, issue #897

* Refactored chat command handling for efficiency. Chat commands no longer sent as CMD_ZERO for authing as this script handles auth so we can auth locally to save a link_message round trip on every chat command. 

*Restored object command handling as per v7.x and previous, using interface channel rather than HUDchannel as there seems no reason to duplicate and hudchannel stuff hasn't  worked for a few years anyway. Added remote auth function -  llSay(g_iInterfaceChannel,"checkauth 1111"); will return "AuthReply|(wearerkey)|(auth level)"  on channel 1111, reporting the auth level of the object owner. Commands can be prefixed with  "authas:(userkey)=", which will use the LOWER auth level between object owner and userkey. 

  Commmand format is targetkey:chat command. Examples: 
 "authas:(userkey)=(targetkey):kneel" - will issue kneel command if (userkey) and object owner both have valid auth "(targetkey):sit (sittarget key)" - sit wearer on (sittarget key) if object owner has valid auth

*menuto cleanup requires menuto target to be in sim AND be the owner of the issuing command

*Set g_iStartup to TRUE in active state on_rez event to restore "owned by" message #906 

*Updated old oc_leash_on_collision "grabby post" script to work with updated interface channel

*Created new oc_smart_post script, a heavily commented example of what can be done via the interface channel

### DIFF
--- a/src/collar/oc_api.lsl
+++ b/src/collar/oc_api.lsl
@@ -1,3 +1,4 @@
+
 /*
 This file is a part of OpenCollar.
 Copyright ©2021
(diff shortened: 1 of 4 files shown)


Medea-Destiny:
Note: the oc_smart_post restrictions function is dependent on the oc_rlvsuite change in pr #1015

Pingout:
Tested by Omnia and Ping. These changes work as expected. 
The # wildcard has been restored, allowing the collar wearing owner to command multiple other collars without affecting themselves. For example the command "# lock" will lock every owned nearby collar, and leave the collar wearing owner's collar unlocked.
"Owned By" message on startup has been restored. The collar reports all owners. Trusted are not reported on startup
In general, commands seem to be handled quicker. We have no metric to prove this, it just feels faster
The Grabby post now works very well. We did report to Medea that a sub who remains in collision with the post will receive multiple rapid leash attempts.
The Smart Post also works well. Medea cleaned up a couple of debug messages that were in the script
Thank you!

SilkieSabra:
I'll wait for the Do Not Merge tag to be removed before merging

NikkiLacrima:
There are some conflicts with oc_api from #1001, if have made a version resolving the differences,  will make some basic inworld checks and a PR later today

NikkiLacrima:
Instead of wrestling with the github based conflict resolution with #1001, I will merge the changes in my own code, do some inworld sanity checks and then create a PR with the combined version of oc_api

Medea-Destiny:
Conflicts with #1001  /#1010 should be fixed. Quick test that command passing  / Safeword / Runaway working as intended with Nikki's fixes needed.

Pingout (approved):
Tested by Omnia and Ping. Everything works as expected

PR #1017, opened 2023-11-29
Update oc_detach.lsl

### LINKED ISSUE #875: Detach chat command fails
8.2.2 and older

chat message {prefix} detach fails due to incorrect filtering of usercommand() function. Only accepts "{prefix} menu Detach" (with capital letter) at the moment

Medea-Destiny:
Fix for filtering of menu command so that both "menu detach" and "menu Detach" will be accepted. fixes #875 

(untested in world)

### DIFF
--- a/src/Apps/oc_detach.lsl
+++ b/src/Apps/oc_detach.lsl
@@ -7,8 +7,9 @@ Copyright ©2021
 
 Aria (Tashia Redrose)
     * Dec 2019      - Rewrote Outfits & Reset Script Version to 1.0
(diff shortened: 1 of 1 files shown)


Pingout (approved):
Tested by Omnia and Ping. This works as advertised :)
The [prefix] Detach, [prefix] detach, and [prefix] DETACH all bring up the Detach App menu

Also tested [prefix] DeTaCh and [prefix] dEtAcH for fun and they worked fine too

PR #1024, opened 2023-12-17
Update oc_undress.lsl

Medea-Destiny:
Making the RLV clear function clear outfit locks


Medea-Destiny:
Problem identified by a friend. Fix seems to work fine but could use an external tester.

Pingout (approved):
This works as expected. The CLEAR ALL command in the RLV menu now removes Clothing Locks. I did not observe any undesired side effects of this change. Thank you.

Issue #1033, opened 2024-04-01
oc_dialog stack heap on sit command.


tayaphidoux:
What version of OpenCollar are you using?
8.2.3

What behavior did you expect?
tmsit or *sit to present my owner with a list of objects

What behavior did you see instead?
stack heap collision every time and likely no or incomplete menu of items

What steps does someone need to take to reproduce the problem?
use sit command or menu in a location with more than 16 items.

Medea-Destiny:
LL just upped the response from llSensor from 16 to 32 items, thus potentially doubling the size of the list we're trying to create.

PR #1034, opened 2024-04-03
Emergency oc_dialog.lsl fix

Medea-Destiny:
quick and dirty fix for #1033 . This just trims the sensor return back to 16 items. Should resolve the issue for now.


Pingout (approved):
I grabbed the fix here and did a couple of quick tests. First of all I noticed i had no problem with Stack/Heap with the sensor getting 32 items. I tried it on a couple of collars and no error. However I went ahead and tested the fix anyway, and it did cut down the sensor reply to 16 items, as advertised. 
Works great, thank you Medea

Issue #1048, opened 2024-06-22
Feature Request collar addon aproval


tayaphidoux:
What version of OpenCollar are you using?
8.2.3

What behavior did you expect?
aproval process to be targeted to prevent cross contamination of startup process of addons

What behavior did you see instead?
i have two or more addons thinking they are connected when another one connects.
[09:56:15 AM PDT]  A.R.C Nanny McFeedoux: [Recived approved]TE - Paws Collar Safra's Pony - chest - test (neck):{"addon_name":"OpenCollar","pkt_type":"approved","iNum":0,"sMsg":"","kID":""}

What steps does someone need to take to reproduce the problem?
should include the addon key and/or name that sent the request since the sMsg and kID fields are blank those are perfect spots.

tayaphidoux:
just applied this to my own collar and it seems to be working [11:07:36 AM PDT]  A.R.C Nanny McFeedoux: [Recived approved]TE - Paws Collar Safra's Pony - chest - test (neck):{"addon_name":"OpenCollar","pkt_type":"approved","iNum":0,"sMsg":"Nanny","kID":"6b024578-e705-1634-5044-374f2c79f66f"}

SilkieSabra:
Please do test it *before* you make the pr or else open it as an issue and include the fix you made there, then test it yourself, and on at least one other person, then make the pr when you're sure to the best of your ability it works.  Please.  For future reference.

tayaphidoux:
thats what this is is an open issue/feature request and i applied the consept jus now to test it.

tayaphidoux:
after i can get some relogs if there is no pr's for it yet i may make one.

SilkieSabra:
Oops sorry.. I was too hasty, I'll remove this section of the thread

SilkieSabra:
Amended---yes please do once you've tested it thoroughly.

tayaphidoux:
oc_addon uses llRegionSayTo and i am being silly thinking it could cross talk.

PR #1049, opened 2024-06-27
upate Addon Aproval message and filtering

tayaphidoux:
Update the addon aproval message and filtering to prevent addon confusion when you have multiple addons and one is aproved but the other has yet to be approved.

This fixes an issue with multiple addons that could in some cases break their connection to the collar.
#1048

### DIFF
--- a/src/collar/oc_addons.lsl
+++ b/src/collar/oc_addons.lsl
@@ -17,7 +17,7 @@ KristenMynx
 Ping (pingout.duffield)
    April 2024            -  Fix at Initial Handshake to disregard any nearby rezzed collar. Ref Issue #1038
 Phidoux (taya.maruti)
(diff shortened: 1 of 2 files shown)


Pingout:
Hi Taya, thank you for this. I'm a little confused about which scripts updated and need testing. Would it be possible to send me the updated scripts inworld, for testing.

tayaphidoux:
> Hi Taya, thank you for this. I'm a little confused about which scripts updated and need testing. Would it be possible to send me the updated scripts inworld, for testing.

oc_addon and oc_addon_template, i tested it with my minicon project which is alot more modified than the template to make a less volitile addon script, but sure i can send you a copy with the modifications.

tayaphidoux:
ignore me i am just being silly i forgot entirely that it uses llRegionSayTo which already eliminates cross talk, i must be to tired to think straight

PR #1050, opened 2024-06-30
Fix oc_folder "function returns void, no value allowed"

SueCripter:
When trying to compile oc_folders.lsl in OpenSim, we get the error "function returns void, no value allowed".

There are multiple offending rows, but they all looks like this: "if(iAuth == CMD_TRUSTED && !Bool((g_iAccessBitSet&1)))**return R()**;"

The function that the code is in ("Browser(key kID, integer iAuth, string sPath)") and the function that have the value returned ("R()") are both no-returners.
In other words, it is incorrect code to return values.
Apparently SL is fine with this invalid code, while OpenSim protests.

To solution is: "if(iAuth == CMD_TRUSTED && !Bool((g_iAccessBitSet&1)))**{ R(); return; }**"

According to my understanding code would definitely not be changed just for OpenSim, but in this case the change needed to have "correct" code is small, there is basically nothing to test, and thus should be able to be considered.


Medea-Destiny:
@SueCripter I've approved code but please add a contributor's comment at the top explaining the change, along the lines:
 ```*Sue Cripter
    June 2024: Changed instances of return R(); to {R(); return;} for OS compatibility.```
    
  Then we're good to merge.

SueCripter:
@Medea-Destiny Does it look good now?

Medea-Destiny (approved):

Medea-Destiny:
@SueCripter Yep, all good! Thanks.

SilkieSabra:
Does this change not need to move to the 8.3 branch? This is why I want to change things so master is the dev branch.

Medea-Destiny:
Wow. Well, rebasing to 8.3 features branch caused Github to complain about a non-existing merge conflict. However in trying to fix this I observe that the same changes were merged into 8.3 back in november, without a credit. Please folks, ensure there is a changelog/credit note at the top of scripts before merging changes! Otherwise things can get missed.

I have done a new pr at #1059 to add a credit note to both Sue and the original PR writer Neil2Ball. Let's merge that and then we no longer need this.

Medea-Destiny:
Handled in other PRs

Issue #1053, opened 2024-07-06
/1 <prefix> NAME displaying incorrectly when set


mistressohm:
**What version of OpenCollar are you using?**
8.2.3000 (Hotfix oc_dialog)

**What behavior did you expect?**
When using /1 _prefix_ Name _New Name_ the response would list _New Name_ in confirmation

**What behavior did you see instead?**

/1 _prefix_ Name _New Name_ responds with 

[device name]: The wearer's name is now set to _Previous Name_

**What steps does someone need to take to reproduce the problem?**

Wear (or have an owned submissive wear) a version 8.2.3000 (Hotfix oc_dialog) OpenCollar

Use the Name chat command:  /1 _prefix_ Name _new name here_

Instead of seeing the new name, you will instead see the prior name in the confirmation.  However, if you grab the wearer's leash, you will see the new name displayed.

To verify, type: 

/1 _prefix_ Name
and you should see the new name, and not the previous name.

I suspect that there is a statement to print the name to local chat as a confirmation, that is getting called BEFORE the routine to actually change the name value, instead of AFTER that value has been changed.

Medea-Destiny:
The link message to trigger the notification was done immediately after the link message to update the name, and was arriving before that was processed. We already have a delay to avoid this in setting device name, but it was missing in name. Have added in the delay and put some clarification into the notification for changing both name and device name in case of lag.

@mistressohm Thanks for the report, small thing but we really do rely on people spotting these small things and letting us know!

PR #999, opened 2023-10-22
Update oc_core.lsl for wearer setting of safeword and device name

Medea-Destiny:
Fixes for #986 and #987. This version of oc_core allows wearer to change own safeword and device name when not owner. Fix includes improvements to notifications to avoid silent authorization failures.

REQUIRES IN-WORLD TESTING before merge:
Wearer should now be able to change safeword and device name, as well as owner.
Wearer should not be able to disable safeword, only owner.
Wearer should now be notified if safeword is disabled or if device name or safeword is changed.
"(prefix) safeword" should inform wearer or owner of current safeword.
"(prefix) device name" should inform wearer or owner of current device name.
User who is neither wearer nor owner should get clear no access notification for safeword or device commands rather than silent failure.

Notes:
Refactor of safeword function in usercommand. 'Safeword off' now no longer sets safeword to 'off' before disabling, resulting in confusing "Safeword is now set to 'off'" message. Instead safeword off is clearly notified. Wearer can now set their own safeword, but only owners can disable it still. See issue # 986. Attempting to access safeword without permission now gives no access response. 
Provide no access notification for device name, and allow non-owner wearer to name. Notify wearer as well when another person changes device name. See issue # 987

PR #1054, opened 2024-07-09
oc_core fix for reporting previous name on name change

Medea-Destiny:
See issue #1053


Medea-Destiny:
Testing: change name, change device name. (/1 [prefix] name [newname] and /1 [prefix] device name [newname]. Names should change, you should get a report of the new name not the old one. Simple one, this. Mainly testing is to ensure I didn't typo and it compiles. :D

Medea-Destiny:
Note that the changes here are already merged into the oc_core script in #999, assuming that one goes ahead, it'll be unnecessary to merge this afterwards.

PR #1055, opened 2024-07-12
Update oc_timer.lsl

Medea-Destiny:
This is a complete rewrite of the old semi-broken oc_timer plugin, that can move it from being something of limited value that didn't entirely work any more to being a far more useful universal timer function for OC. Using this, it's now possible to unlock virtually ANY restrictions, all together or individually, on a timer. Alternatively it can be used to implement restrictions or functions on a time delay.

It has built in functions to clear restirctions, unleash and unpose on timer, but also allows a list of custom commands to be entered in the form of chat commands separated by a collar that will be executed when the timer runs out. Thus it could be used to for example send someone home after the time is up, or to free them from a cage on the timer, or to unlock a folder that has been locked on them and detach the attachments within it on a timer, or virtually anything else. 

It is written using LinksetData using the methods that will be used in OC9.x. However it's stand-alone and runs on OC8.3. There are still a number of functions it will not use LSD for, notably the same old auth/dialog system via link messages, rather than handling auth locally which will be possible when OC_API and OC_Settings are migrated to LSD.

This PR includes changes to oc_api to re-enable a wearer lockout function. This is something the wearer has to allow, and when enabled, the wearer is locked out of their own collar while the timer is active. This enables using a self-owned OC for self-bondage. Note that this is an extensible function, not unique to timer, and could be used for other functions in future such as capture, without interfering.

### DIFF
--- a/src/Apps/oc_timer.lsl
+++ b/src/Apps/oc_timer.lsl
@@ -1,580 +1,495 @@
-// This file is part of OpenCollar.
-// Copyright (c) 2008 - 2017 Satomi Ahn, Nandana Singh, Joy Stipe,     
-// Wendy Starfall, Master Starship, Medea Destiny, littlemousy,      
-// Romka Swallowtail, Sumi Perl, Keiyra Aeon, Garvin Twine et al.     
-// Licensed under the GPLv2.  See LICENSE for full details. 
+/*
+This file is a part of OpenCollar.
+Created by Medea Destiny
+Copyright ©2024 
 
-string g_sAppVersion = "¹⋅⁴";
+: Contributors :
 
-string g_sSubMenu = "Timer";
-string g_sParentMenu = "Apps";
+Medea (Medea Destiny)
+    *May 2024   -   Created script                             
+et al.
+Licensed under the GPLv2. See LICENSE for full details.
+https://github.com/OpenCollarTeam/OpenCollar
+*/
 
-//MESSAGE MAP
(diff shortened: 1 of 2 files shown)


Medea-Destiny:
Updated script to have:
An unsit function that will unsit the avatar, respecting strict sit (i.e if timer set by someone with insufficient auth, it won't unsit them, while if timer setting has sufficient auth, it will
A  floating text title showing time left (rounded to nearest minute) that works nicely with the titler app, i.e it will not override a titler title if one is set, and will take text colour and title offset settings from the titler app.

Pingout (approved):
So yea, we did a bunch of testing on this, and I'm happy to say it worked well on each occasion. At first I thought there might be a problem with the time showing in the Title, but when I re-read the instructions I could see this was acting correctly, that when a title is set, the Timer does not interfere, so time remaining doesn't show. 
Other than that everything worked amazingly well, with a long timer not losing time
So:
Core functionality. Menus and chat commands working as expected  Yes
Does the "timer time ????" chat command manage to understand varied input well?  Yes
Does timer behave as expected with safeword / runaway?  Safeword: yes / Runaway: 
Does the timer last the right amount of time within reasonable expected variations? Yes
Does logging out and back in some time later run smoothly?  Yes

Auth and permissive work as expected

Pingout:
I noticed a bug in the Title countdown. When the Timer reaches zero, the Title does not clear itself, instead it shows "Timer: 0s Remaining"

PR #1068, opened 2024-08-01
oc_bookmarks fix broken links

Medea-Destiny:
Escapes region names in SLURL reported to collar user when teleporting collar wearer. Unescaped URLs were breaking the slurl if the sim has a multi-word name, by simply escaping the sim name with llEscape we get a functional SLURL.


SilkieSabra (approved):
This worked on our test.  It needs another approving review.

Pingout (approved):
This appears to work as intended. The script now generates the Menu Button correctly and then teleports the user upon clicking the desired button.
Adding and removing Destinations worked well.

PR #1069, opened 2024-08-03
Add Force TP to realleash

Medea-Destiny:
This fix takes advantage of the EXC_REFRESH function to add ForceTP to strict leash, so that when strict leash is active, a leasher can send a TP lure to leashee and force tp them regardless of whether they normally have Force TP ability. When the strict leash restrictions end (after leash is dropped, or 60 seconds after leash holder goes out of range), the ForceTP along with other strict leash restrictions will be dropped, but existing ForceTP set through exceptions will not be impacted.

### DIFF
--- a/src/collar/oc_leash.lsl
+++ b/src/collar/oc_leash.lsl
@@ -1,6 +1,6 @@
 
 /* This file is part of OpenCollar.
- Copyright (c) 2008 - 2016 Nandana Singh, Lulu Pink, Garvin Twine,    
+ Copyright (c) 2008 - 2024 Nandana Singh, Lulu Pink, Garvin Twine,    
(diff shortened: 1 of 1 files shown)


Pingout (approved):
Tested with Trinkitz. We checked the use of this change for many different configuration of Access, everything from Trusted, Captured, non-Owner, Unowned, and no access. Each scenario worked perfectly.

SilkieSabra (approved):
Trinkitz and I tested this.  It works as expected.  It does add some power to public access since.. if someone leashes the wearer on public access, and then an owner teleports them, while the leash is still active the public access user can teleport them back.  This might be fun but also might surprise the owner.

PR #1074, opened 2024-08-11
Update all version numbers for 8.3

Medea-Destiny:


SilkieSabra:
I reviewed each of these version lines and they are correct.

PR #1082, opened 2024-08-20
Change legacy permissions notification in oc_couples

Medea-Destiny:
Message " Inviting (couple partner) to a couples animation using legacy permissions." when not using autoacc might be confusing. Just remove "using legacy permissions".

### DIFF
--- a/src/collar/oc_couples.lsl
+++ b/src/collar/oc_couples.lsl
@@ -643,7 +643,7 @@ state active
 			llSetTimerEvent(0.0);
 			timer_reason = "";
 			ping_sent_to_key = NULL_KEY;
-			llMessageLinked(LINK_SET,NOTIFY,"0"+"Inviting "+ g_sPartnerName + " to a couples animation using legacy permissions.",g_kWearer);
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
OK, let's have a quick discussion on this, as it does remove useful info. Better wording suggestions? Perhaps something along the lines "Inviting (partner) to couples animation and awaiting them to accept animate permission"?

SilkieSabra:
Honestly my feeling is that the behavior where the person with collar access accepts an animation  permission request from the wearer is the existing configuration and does not need to have its text changed.  "Inviting (partner) to a couples animation" already  implies waiting for consent.  That much is fine, and additional information about what *would* happen if you had x button checked instead, does not further anything.

Pingout:
I feel less chatty is better.

SilkieSabra:
Another thought on this one.. the couples line is in the domain of rp, mixing in text speak is jarring.

SilkieSabra:
My pr was intended to resolve the discussion since there was no response to three posts.  Please explain why you feel this change to the message users see when they issue a couples invitation is a necessary change.

PR #1084, opened 2024-09-01
Update oc_couples.lsl

SilkieSabra:
Because users will see this text and will not understand what "legacy" means here, and the change is unnecessary.


Medea-Destiny:
This is a duplicate of #1082

PR #1087, opened 2024-09-05
Update oc_particle.lsl

Medea-Destiny:
Random error message being chatted in local. We have various debug messages coming through in the oc_particle script for lockguard commands being whispered, with the llSay commands normally commented out. I assume this is just a miss on commenting out a debug message, so commenting this one out. There's no real value in this error message other than debug

I'm slightly concerned that the lockguard command was being processed at all, because as far as I can see the collar was reacting to a lockguard message from someone else's attachment. Shouldn't this be filtered on avatar target @NikkiLacrima ? However this will avoid random spamming of the message

### DIFF
--- a/src/collar/oc_particle.lsl
+++ b/src/collar/oc_particle.lsl
@@ -806,7 +806,7 @@ state active
                     }  else if (sLGCMD == "ping") {
                         llWhisper( g_iChan_LOCKGUARD, "lockguard " + (string)llGetOwner() + " " +  sLGPoint + " okay" );
                     }  else {
-                        llWhisper(0, "Unknown LockGuard command: "+sLGCMD);
+                        //llWhisper(0, "Unknown LockGuard command: "+sLGCMD);
                     }   
                 }
                 if (iIsLinking) {


NikkiLacrima:
Agreed the error message can be commented out, its there since some furniture sends misformed LockGuard commands. Also the collar should check that the commands are intended for the wearer, as is done by the cuff scripts. The reason this has not caused a lot of issues is probably, my guess, that most collars dont have the side rings that attaches chains and most funiture dont tries to link to the collar points.

Anyway something like the following could be added at line 752:
```    listen(integer iChannel, string sName, key kID, string sMessage) {
        if (iChannel == g_iChan_LOCKGUARD){
            // Implementation of the Lockguard V2 Protocol
            list lLGCmd = llParseString2List(llToLower(sMessage), [" "],[]);
            if (llList2String(lLGCmd,0) == "lockguard") {
                key kLGAv = llList2Key(lLGCmd,1);           // Request Avatar-UUID
+++                if (kLGAv != g_kWearer) {                    // Check that wearer is target
+++                   return; 
+++                }
                string sLGPoint = llList2String(lLGCmd,2);  // Request ChainPoint
```

NikkiLacrima:
Code looks good, can be committed

NikkiLacrima (approved):
Looks good, ready to merge

Medea-Destiny:
Superseded by @NikkiLacrima's complete fix at #1088, closing.

PR #1088, opened 2024-09-05
Nikki lacrima patch particles

NikkiLacrima:
The LockGuard handler in oc_particles should check that the LG message is actually intended for the wearer, follow up on #1087


Medea-Destiny (approved):
Looks good to me!

Medea-Destiny:
Addresses the filtering point I raised in #1087, and includes the typo fix for commenting out of the debug message. Closing #1087 as this supersedes and it's @NikkiLacrima's baby. Thanks for the quick response Nikki!

PR #1091, opened 2024-09-07
particle leashlength fix

NikkiLacrima:
Read the setting "leash_leashlength" and not incorrectly "leash_length"


Medea-Destiny (changes requested):
Good spot! 
Can you insert "Sept 2024: Fix setting name for leash length" in line 10 please?

Medea-Destiny (approved):
Thanks Nikki!

PR #1093, opened 2024-09-09
Nikki lacrima minorclean rlvsuite

NikkiLacrima:
Two are parts:
Remove superflous llOwnerSay and change the "0" flag to "1" in the NOTIFY when modifying restrictions.
The current code sends two sets of notifications to wearer if wearer changes restrictions.

g_iMenuStride was unitintialized so lets remove kMenuID stuff and replace with sName+"~"+llGetScriptName() as in the #963 patch. This change has been in oc_api and oc_core for quite a while now so should be considered well tested. Code is simpler and uses slightly less memory.


Medea-Destiny:
Looks like good and sensible fixes to me. However as we've got release candidates for 8.3 out I'd suggest it might be best to push this to 8.4 for a release a few months down the line as these are efficiency improvements rather than bugfixes and we really want to get 8.3 actually release :D Thoughts?

NikkiLacrima:
These changes are not critical, just minor annoyances. So they can wait, but if we prepare another RC before release of 8.3 they might go in.

Medea-Destiny (approved):
Code looks fine

PR #1096, opened 2024-09-13
OBJECT_TEMP_ATTACHED rejection in oc_API for interface channel

Medea-Destiny:
Scottie Muircastle just pointed out to me a potential vulnerability in the object interface function. As temporary attachments become "owned" by the wearer, in theory people could use a temporary attachment that's attached as a furniture prop to send OC command with the collar wearer's auth. This was an issue previously with RLVa rejecting RLV commands from temp attachments as it's a potential vector for bypassing auth.

I've chucked in an extra line to simply reject arbitary commands from temp attachments which should resolve this. We should consider having this as a user setting or similar. I haven't had time to test this as yet but as it's a potential security flaw we need to consider this as a security patch for 8.3

### DIFF
--- a/src/collar/oc_api.lsl
+++ b/src/collar/oc_api.lsl
@@ -529,6 +529,7 @@ state active
                 return;
             }
             else {
+                if(llList2Integer(llGetObjectDetails(i,[OBJECT_TEMP_ATTACHED])==1) return;
                 key kAuthKey=llGetOwnerKey(i);
                 integer iAuth=CalcAuth(kAuthKey);
                 if(llGetSubString(m,0,6)=="authas:"){ //messages prefixed authas:(key)=(cmd) will use the auth level of key if LOWER than object owner auth.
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
For ease of testing I've made an object you can attach manually and it'll use the interface channel to make you kneel. However if you rez it and click it, it'll ask permission to temp attach, and then you won't kneel. Ask me in world if you want a copy for testing.

Medea-Destiny:
We could consider if an object is a temp attachment, adding a check for OBJECT_REZZER_KEY and allowing it to pass auth based on the rezzing object, but this will take a little consideration so I recommend targeting 8.4 for that.

NikkiLacrima:
If I understand the code right it rejects temp attachment both for wearer and for other avatars ??

Medea-Destiny:
Right. The point here is that a temp-attached item will have the wearer as owner, and the object auth relies on the object owner for primary auth. The potential risk scenario here is that for example someone could sit on an item of furniture that temp-attaches an object, potentially using experience permissions (avsitter experiencewill do this for example). That item then inherits the wearer's auth level even though as a temp attachment it's not something the wearer necessarily has any knowledge of what it's doing. 

This patch rejects anything that's temp attached, which makes sense because another avatar with auth permission on a collar could be the vector of such an exploit.

NikkiLacrima (commented):
Looks good

Pingout:
Very good. It's some what of a convoluted volnerability, but we can't leave it. Thanks for the fix.

Issue #1095, opened 2024-09-13
Stale auth permitting Force Sit to continue to be used by user on public, after public switched off


Medea-Destiny:
To reproduce:
Have public access on someone's collar.
Go to RLV>force sit, get list of objects wearer can be sat on
Wearer switches off public access
Select an item to sit them on. Sit will be successful, even though public access is now off.
Wait for menu to reappear, and repeat ad nauseum.

The quick fix: change line 878 (OC 8.3) of oc_rlvextension to ```llMessageLinked(LINK_THIS,0,"sit",(key)llList2String(to,1));``` -- this will re-auth the user before generating a new force sit menu, rather than using the "current" auth.

The bigger issue: This is a by-product of remenuing using the auth value originally assigned. We have no way of indicating that auth has changed. Most of the time this isn't going to matter, but in remenuing within a script, we're generally just feeding the current auth back in to the system. The force sit menu is only one more obvious place where this is happening, but similar things are happening all over the place. This could have been addressed with an AUTH_STALE flag that would be sent on any change to auths, which would tell scripts to ignore any "saved" auth level (part of my proposal for OC4, but that's ancient history...) but this requires a lot of work throughout the collar.

Going forwards, we're going to use linkset Data for auth levels. This will allow every script to check the current auth without having to route back through linked messages, and will mean that the current auth level is always respected. While we might want to drop this fix in for force sit in the short term (8.4?), ultimately LSD will fix all these types of problems.

NikkiLacrima:
(Untested) Since rlvextensions have the buggy noninitalized g_iMenuStride it should be enough for the wearer to simply open any menu from rlvextension, this will invalidate the other users menu.

Seriously, its better to start getting the LSD framework set up, perhaps for only global_auth at first.

Medea-Destiny:
Yes absolutely the plan to start moving auth and a few other things to LSD in 8.4, while maintaining the 8.x api (i.e oc_api will still handle the auth process but checkauth function there will refer to LSD values rather than script variables, and we can auth locally on an ad-hoc basis where it seems particularly necessary or efficient). 

Smart thinking on clearing the dialog list as an interim fix though.

PR #1097, opened 2024-09-13
Clear menus to invalidate auth when public mode is turned off

NikkiLacrima:
Listening to LM_SETTING_EMPTY for "auth_public".
Uses RemoveMenuStride() to close menu listeners.

Should solve issue #1095

Basic inworld testing done and works so far.

HOW TO TEST:

Check that all normal collar menus still are working ok, with owner and a friend accessing the menus.

Set collar to public and let a co tester who is not owner try that RLV force sit works.
Then let the friend open a RLV/Forcesit meny and just keep it open.
Now change to non public, and after that the friend makes a force sit selection.
* This should now have no effect at all, and friend have to reopen menu

For comparison try the same on a basic 8.2.3 collar


Medea-Destiny:
Smart solution! How about we extend this with ```if(sStr=="auth_public" || sStr=="auth_group")``` to catch the same issue if authed on group and group access is removed? 

For 8.4 with LSD auth levels, we can auth locally in oc_dialog listen event to provide fresh auth on each use of dialog. This current  fix will mean trusted users lose active menus on changing public access which is a touch weird but a small price to pay to patch a security loophole in 8.3.

Medea-Destiny (approved):
Code looks right to me

Pingout:
Looks great.

PR #1117, opened 2024-10-19
Revert "g_iAwayCounter clearing restrictions when nothing has changed"

SilkieSabra:
Reverts OpenCollarTeam/OpenCollar#1110  Surely this is minor and just needs a quick code review.  I need more coffee.


NikkiLacrima:
I suggest the merged #1110 is tested and then the revert can be cancelled (reverted). Same with the other 2.
I have them running in my adventurous alts' collar and no ill effects have been seen. If any are found we can patch them instead.

/Nikki

Issue #1119, opened 2024-11-14
OC_dialog crashing in V 8.3 when detach is attempted


SilkieSabra:
In V 8.3, a user wearing quite a few items with long names is crashing oc_dialog when someone tries to open the detach menu from Apps.  The rest of the collar is still able to be used.  The Detach button in apps stops producing the detach menu and oc_dialog throws the error.
https://gyazo.com/2babcc6899f086e1a75ffedf1d7744dc 
I don't know why the script error is coming from oc_dialog but no doubt there's an explanation -- or why the rest of the collar continues to work after a stack heap message coming from oc_dialog.

Do we need to look at oc_detach?

Here's the list of what this user was wearing.  Reducing the number of worn items allowed oc_detach to work and did not produce a script error from oc_dialog.

*PKC* Queen Bee Bento Rings - Legacy Left {Creator} [R hand] 
 *PKC* Queen Bee Bento Rings - Legacy Right {Creator} [R hand] 
  / EYES / lel EvoX 3.1 {Creator} [L eye] 
  / HEAD / lel EvoX SIWA 3.1 {Creator} [chin] 
 / TEETH / LeLUTKA 3.1 {Creator} [] 
  AVEC TOI - Wendy O Stilettos (Legacy) {Creator} [pelvis] 
  AVEC TOI - Wendy O Stockings (Legacy) {Creator} [pelvis] 
 Cynful Mina Piercings - Back Dimples L - Legacy F {Creator} [R pectoral] 
  Cynful Mina Piercings - Back Dimples R - Legacy F {Creator} [L pectoral] 
  Cynful Mina Piercings - Belly - Legacy Female {Creator} [stomach] 
 Cynful Mina Piercings - Cleavage - Legacy {Creator} [chest] 
  Cynful Mina Piercings - Collarbone - Legacy Female {Creator} [] 
  Cynful Mina Piercings - Spine - Legacy F {Creator} [back] 
  It's NOT Mine! Complete Body Cum System {Creator} [pelvis] 
  LeLUTKA.Png.SIWA.Ear.Set {Creator} [R ear] 
  MIDNA - Alice Piercing - GIFT {Creator} [L ear] 
 Piercing de nez or et diamant {Creator} [nose] 
 RAWR! Delerium Rings Legacy {Creator} [nose] 
 RAWR! Plumb Armlets Legacy {Creator} [R upper arm] 
  RAWR! Plumb Necklace Legacy {Creator} [neck] 
[...]

PR #1126, opened 2025-01-24
Improved memory handling in oc_dialog::SortUUIDList

NikkiLacrima:
Improved memory handling and early release of allocated resources should improve situation for Issue #1119. Simple drop in replacement for oc_dialog. 

This version has been tested for two months and no regressions seen. 

As long as only a single owner is accessing the Detach app menus at a time problem should be fixed.

### DIFF
--- a/src/collar/oc_dialog.lsl
+++ b/src/collar/oc_dialog.lsl
@@ -17,6 +17,7 @@ Medea Destiny -
 Nikki Lacrima
     Aug 2023  - Changed functions for clearing auth so  auth doesn't persist for open menus
     Sep 2024  - Invalidate all menus except wearer when public or group mode is turned off
+    Jan 2025  - Improved memory handling in SortUUIDList, clear resources as early as possible
 
 */
 integer CMD_ZERO = 0;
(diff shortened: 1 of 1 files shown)


NikkiLacrima:
This patch also allow the full number of 32 sensor results to be used in the sensor handling, not in the current patch but its a simple change. Also tested in places where force sit returns 32 results.

Pingout (approved):
Tested for functionality, everything appears to work as expected. The Force Sit function now returns 32 items without error, and the detach app will also return 32 worn items without crashing.
I believe that @Medea-Destiny would like to review this scrpt for workflow. I will approve this and tag it with Don't Merge Yet awaiting Medea's review

Medea-Destiny:
Looks great at a first go through, but I'll take a closer look in case I can think of any possible issues / interactions. For the sake of safety (and possibly menu legibility) it might be an idea to continue to clamp the force sit function  - maybe to 20 instead of 16? How do people feel about that?

NikkiLacrima:
The issue with 32 force sits is actually the same as with 32 worn items in Detach, so why ??

NikkiLacrima:
Maybe its about time to either accept this patch or tell whats wrong, another duplicate of the issue and this fix has been known for 4 months. #1134 duplicates #1119.

Slow and careful is good , but slow and perhaps look at things another month is a good reason to fail

Medea-Destiny (approved):

Medea-Destiny:
Looks good, let's go ahead and merge this.

SilkieSabra:
Ok Medea says let's merge but there's still a Do Not Merge Yet label on it and it says the branch is out of date, do I press Update branch and then merge or what?

PR #1128, opened 2025-02-26
Update oc_particle.lsl

NikkiLacrima:
Add support for multiple simultaneous leashpoints on the collar.

The background is to be able to leash to a pony bit with two reins, one to each bit ring.
The "leashpoint" prim is moved to one on the bit rings.
A new prim named "fcollar" is is added at the other bit ring.

Having two prims named like this will create two entries in the g_lLeashPrims list with name "collar", one entry for each of these two prims.

The script will generate two chains of leash leash particles, one from each of the two leashprims ending at the leash holder.

### DIFF
--- a/src/collar/oc_particle.lsl
+++ b/src/collar/oc_particle.lsl
@@ -8,6 +8,7 @@
  Nikki Lacrima
     Aug 2023: Updated for lockguard chain texture
     Sept 2024: Fix setting name for leash length
+    Feb 2025: Support for two leashpoints, "leashpoint" and "fcollar", for ponybits
     
 */
 string g_sScriptVersion = "8.3";
@@ -177,6 +178,7 @@ FindLinkedPrims() {
     if (llListFindList(g_lLeashPrims,["fcollar"]) < 0){
         g_lLeashPrims += ["fcollar",LINK_ROOT, "collar", LINK_ROOT, "collarfrontloop", LINK_ROOT];
     }
+    //llOwnerSay(llDumpList2String(g_lLeashPrims," "));
 }
 
 Particles(integer iLink, key kParticleTarget, vector vScale) {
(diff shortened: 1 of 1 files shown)


SilkieSabra:
This could have multiple other applications with such a simple fix ! We've been looking for how to have multiple leash points for ages.

Pingout:
I tested this with a couple of collars, and it worked great. I do have one question, though: Is "fcollar" still going to be a valid chainpoint in the Addon Cuffs system? Could we name this prim something besides "fcollar"?
Please see https://opencollar.cc/docs/OpenCollar-Cuffs, the section on Collar Chainpoints

NikkiLacrima:
It is a quick, and simple, fix and we can polish it as we like.

Yes "fcollar" is available as a chainpoint, the current behaviour is that  it is the prim with the lowest link number that is named either "leashpoint","ooc" or "fcollar", and it is reachable by any of the names "fcollar","collar" or "collarfrontloop". 

All linked prims with name  "leashpoint","ooc" or "fcollar" will be used as leash point(s).

The standard collars uses the name "leashpoint" for the "fcollar" chainpoint, but "ooc" or "fcollar" should work as well for a single leashpoint.

A slightly more refined scheme is:
1) If there is a "fcollar" prim: then this prim is used for chainpoints "fcollar","collar" and "collarfrontloop", and any prim named "leashpoint" or "ooc" is used as multiple leashpoint. Only one prim named "fcollar" will be used.

I2) f there is no "fcollar" prim: then the first prim named "leashpoint" or "ooc" is used for chainpoints "fcollar","collar" and "collarfrontloop", and any further prim named "leashpoint" or "ooc" is used as multiple leashpoint.

Medea-Destiny (approved):
Here's a code approval

Medea-Destiny:
Code looks good. It's now possible to add extra leash points by naming them leashpointx, and when leashed in the normal manner, multiple leashes will be drawn from both the main leashpoint and any leashpointx prims to the leash target. An fchain link can be added as a front collar leashing point separate to the standard leashpoint link for use with LM/LG furniture, so that it's now possible to have a front collar leashpoint connected to furniture without interfering with use of the leash.

Medea-Destiny (changes requested):
Oh wait! One tiny change required, line 776 has a debug llOwnerSay that needs commenting out.

PR #1131, opened 2025-04-02
Update oc_relay.lsl

Medea-Destiny:
Fixing another instance of the old CMD_WEARER failing when wearer is set to trusted permission

### DIFF
--- a/src/collar/oc_relay.lsl
+++ b/src/collar/oc_relay.lsl
@@ -25,6 +25,10 @@ Nikki Lacrima
                - implemented Yosty7b3's menu streamlining, see pr#963 
 et al.
 
+Medea Destiny
+   *Apr 2025   - Fixed a wearer no access when has trusted permission issue, bug
+               found by Annie Zehetbauer in world.
+
 Licensed under the GPLv2. See LICENSE for full details.
 https://github.com/OpenCollarTeam/OpenCollar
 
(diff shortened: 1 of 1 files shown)


NikkiLacrima:
Tested with settings:
wearer not owner
allow wearer access to relay settings

before patch 

- set wearer not in trusted list, check that wearer can access relay settings

- set wearer trusted in trusted list, check that wearer can not access relay settings

this means that setting wearer to trusted blocks access to relay settings

after patch 

- set wearer not in trusted list, check that wearer can access relay settings

- set wearer trusted in trusted list, check that wearer can still access relay settings

So the patch allows setting wearer to trusted without interfering with access to relay settings.

Looks good and nothing weird in code, approved.

It seems that the auth level CMD_WEARER is badly named, it should probably be someting like CMD_WEARER_DEFAULT_ACCESS since it is not good for testing if the command is from wearer or not

NikkiLacrima (approved):
Patched tested and approved

Medea-Destiny:
> It seems that the auth level CMD_WEARER is badly named, it should probably be someting like CMD_WEARER_DEFAULT_ACCESS since it is not good for testing if the command is from wearer or not

Allowing wearers to set themselves as trusted is what has caused this particular bit of confusion -- before that CMD_WEARER or CMD_OWNER would cover the bases. Where we want to check for wearer id or owner permission, it's probably best to do just do that (as with this particular change).

Potentially as we change over to the LSD auth method we could use a concatenated AUTH system... something like AUTH~OWNER~WEARER, AUTH~TRUSTED~WEARER, AUTH~WEARER etc. but I can't off the top of my head see any other context that's particularly useful and so I think so long as we're checking for id rather than auth value to check for wearer operations specifically, that's probably the simplest and most efficient path?

PR #1147, opened 2025-06-07
Reduce oc_rlvextension.lsl memory usage

mcgroarty:
I want to add new functionality to oc_rlvextension.ls in a subsequent commit. Since that would be built on top of this, this is a separate PR to reduce script memory usage with no functional changes.

- Muffle code refactored
- Some strings shortened
- Removed redundant llParseString2List calls in UserCommand
 
Startup memory shows a reduction of ~1,982 bytes

### DIFF
--- a/src/collar/oc_rlvextension.lsl
+++ b/src/collar/oc_rlvextension.lsl
@@ -68,7 +68,10 @@ string g_sSubMenu1 = "Force Sit";
 string g_sSubMenu2 = "Exceptions";
 integer g_iStrictSit=FALSE; // Default - do not use strict mode
 
-//MESSAGE MAP
+// MESSAGE MAP
+// Authorization constants: a smaller value means a higher privilege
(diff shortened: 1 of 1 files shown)


Medea-Destiny (changes requested):
Thanks for doing the llReplaceSubString change and the improved string contractions, looks good!
Suggestions:
 MenuSetExceptions menutext should use "Force TP - When on, auto-accept their TP offers"


MenuForceSit: Add " except via [UNSIT]" to end of dialog menu text and change "during" to "from"

For the exceptions notify LM on line 548, "Add selected values together for the bitmask value" instead of "Add selected values for the bitmask" 

Minor thing, but can we use a different variable than iPos at line 615? It shouldn't collide with separate declarations of the same variable name in deeper level code blocks later in scope,  but it would save potential reading confusion. 

All the expanded comments are great! We need more of that in general.

PR #1150, opened 2025-06-09
PBR fix  for hide/show, move function out of themes

Medea-Destiny:
Allows Hide/Show to work with PBR materials. This uses a function that's bugged and will not properly revert alpha values correctly (bug reported to LL) so currently it toggles alpha values between 0 and 1. This is how the current hide/show works with BP textures anyway so no big issue, but in future when the llSetGLTFOverrides function is fixed, it will be possible to preserve alpha values for better functionality.

The hide/show has been moved to the states script and excised from the oc_themes script. This means that for collars that do not use themes, it's no longer necessary to have that script just to handle show/hide functionality.

### DIFF
--- a/src/Apps/oc_bell
+++ b/src/Apps/oc_bell
@@ -1,0 +1,523 @@
+// This file is part of OpenCollar.
+// Copyright (c) 2009 - 2024 Cleo Collins, Nandana Singh, Satomi Ahn,   
+// Joy Stipe, Wendy Starfall, Medea Destiny, littlemousy,         
+// Romka Swallowtail, Garvin Twine et al. 
+// Last updated: Trix (TrixAeo) 2024 - Uploaded new sounds.
+// Licensed under the GPLv2.  See LICENSE for full details. 
+
+
+//scans for sounds starting with: bell_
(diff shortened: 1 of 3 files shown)


Medea-Destiny:
Note about linting failure:

This script fails the LSLint check with the following errors:
[>>] Linting src/collar/oc_states.lsl...
ERROR:: ( 89,  9): `llSetLinkGLTFOverrides' is undeclared.
ERROR:: ( 89, 48): `OVERRIDE_GLTF_BASE_ALPHA_MODE' is undeclared.
ERROR:: ( 89,105): `OVERRIDE_GLTF_BASE_ALPHA' is undeclared.
ERROR:: ( 93,  9): `llSetLinkGLTFOverrides' is undeclared.
ERROR:: ( 93, 48): `OVERRIDE_GLTF_BASE_ALPHA_MODE' is undeclared.
ERROR:: ( 93, 82): `OVERRIDE_GLTF_BASE_ALPHA' is undeclared.

This is a new LL script function that just hasn't made it into the LSLint software yet, so it's not being recognized and is being falsely flagged.

PR #1158, opened 2025-07-15
Create tpt.lsl

Trix7125:
First version of  Anne Skydance plugin


Medea-Destiny (changes requested):
Hi @Trix7125 

Please take a look at https://github.com/OpenCollarTeam/OpenCollar/blob/main/src/spares/oc_plugin_template.lsl where we have a template for plugins which you can use to convert this to the standard plugin format. It'll show you how you can utilise the OpenCollar auth system, menus and settings for this project.

PR #1160, opened 2025-07-23
Multirelay and LSD update for oc_relay.lsl

NikkiLacrima:
First version, tested with 4 simultaneous RLV sources and pingpong on relog.

Uses LSD to store restrictions per UUID
Uses LSD for mode and settings

### DIFF
--- a/src/collar/oc_core.lsl
+++ b/src/collar/oc_core.lsl
@@ -43,12 +43,16 @@ Medea (Medea Destiny)
                 -   added delay after name change to ensure report is correct and added clarification text here
                     and in device name. Issue #1053
    Jun 2025     -   Restored HELP button to help/about menu to deliver help notecard                         
+
 Stormed Darkshade (StormedStormy)
     March 2022  -   Added a button for reboot to help/about menu.  
 
 Yosty7B3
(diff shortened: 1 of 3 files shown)


NikkiLacrima:
Update Aug 25:
Handles multiple pending ask sources
Code cleanup

MinesweeperM:
Some trouble with capturing devices on relog.
On relog, relay returns av to device, where it was captured, then device trying to connect relay, but it does not answer, until main collar scripts load. So common devices (like AV_Sitter) decides that the relay is missing, and restrictions does not apply.

NikkiLacrima:
Changes 251018: Improved memory handling in relay, faster and simpler startup.

Hopefully this should be almost final version, still some debug output left.

Medea-Destiny (changes requested):
rlvsys and relay both still have script profiler active in menu (lines 132, 143 & 145 in rlvsys, 167, 177 and 179 in relay).  We might want to leave those in while we're putting this through beta, but should be commented out before merge.

Medea-Destiny (approved):
Code looks in a really solid state to me, I can't think of anything more!

PR #1166, opened 2025-10-17
oc_anim add groundsit function

Medea-Destiny:
New button triggers groundsit. This is a "parking" type feature which will sit the wearer on the ground with standard groundsit animation (from AO) and lock standing. Same or higher auth as the person issuing the command is required to undo and allow moving again.

### DIFF
--- a/src/collar/oc_anim.lsl
+++ b/src/collar/oc_anim.lsl
@@ -289,7 +289,7 @@ UserCommand(integer iNum, string sStr, key kID) {
                     
                 }
             }else llMessageLinked(LINK_SET, NOTIFY, "0%NOACCESS% to changing pose offset", kID);
-        } else if( sChangetype=="groundsit" || sChangeType=="park") {
+        } else if( sChangetype=="groundsit" || sChangetype=="park") {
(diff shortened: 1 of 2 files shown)


NikkiLacrima:
It works as is it should, but using the Unsit from RLV/Force Sit will unsit the wearer while the groundsit still thinks wearer is seated on ground.

NikkiLacrima:
The problem is not that [UNSIT] can force a stand, its rather that this should also cancel the Ground Sit state.

Without to much hard thinkining about pros and cons, maybe Ground Sit should go with Force Sit in that module.

Medea-Destiny:
@NikkiLacrima Ok, updated to work that way. UNSIT will now also trigger groundsit off.

SilkieSabra (commented):
Ok, I found the groundsit button in the Animations menu where I should have looked in the first place.  When I push groundsit there it blocks stand. Strict Sit is not enforced.  To unsit I then go to the Force Sit menu to press Unsit.  Is this the way it's supposed to work with Strict Sit not enforced?

Medea-Destiny:
Made it "Park"

PR #1167, opened 2025-10-17
oc_dialog use LSD

Medea-Destiny:
Extending Nikki's optimizations with use of LSD temp storage of menu buttons to avoid stack/heaps caused by multiple large menus.  Efficiency update to substitutions DIALOG_RENDER_REQ to limit sending menus to addons only when called from an addoin


Medea-Destiny:
@NikkiLacrima Oooh how did I miss the existence of that command? Nice catch Nikki! Very useful. Have updated the script as your recommendation.

PR #1171, opened 2025-11-03
oc_leashaddon - replacement leash point as a collar addon

NikkiLacrima:
Replaces collar leashpoint

Collar still handles leashing, unleashing,  physics and leash settings. The addon leashpoint only redirects the particle chain from collar to the addon prim. 

Menu with Active(on/off), Lock(on/off) and AutoHide(on/of). The leashpoint lock is released once the leashpoint loses contact with the collar, this can take 5 minutes.

To use it simply drop the script into something like a septum ring and then wear it while the collar is used. The script will then connect as an addon for the collar. For detach functions, and autohide to work the script should be added to the root prim. If the intended leash prim is not the root, simply rename to “leashpoint”.


Medea-Destiny (approved):
Have been working with this code in world for a while (and done a few tweaks!), it's good, and a great addition to OC. Has been tested in world by myself and Clover -- Clover's comments: "We used oc_leashaddon in an object by itself and as part of a larger object with a linked "leashpoint" prim and it worked perfectly each time. It even knew to let the leash go back to my collar automagically  if I detached the scripted object while it had the leash drawn to it."

I'm happy with this code and we're going to put this out into R&D for general testing now. Doesn't require any code changes to the collar to work, so any v8 collar will do with it. 

We should think about what the best name for this is, and put some instructions at the top of the script.  Perhaps we should provide this at the HQ with a physical object as an example?

Medea-Destiny:
@SilkieSabra Nothing so much wrong as that it may not fully convey the range of uses it can have. While primarily it's exactly that, an add-on leash point, it will also bring up the collar's menu when clicked. So all those people who want to wear two collars because their main one doesn't match their outfit, all the people who want to wear two collars because they want to be leashed by some alternative body part, are equally catered to. So maybe there's a name that better conveys that to people?  I dunno. :D

Pingout:
We do need instructions. I tried to generate an addon using just the conversation here and I get 2 leashes, one from the collar and another from the nose ring addon i used

NikkiLacrima:
Naming scheme is a bit more complex actually:

The default is to have g_sAddon = "PRIMNAME"; in this case the name of the prim is used as addon name in the addon menu.

Setting g_sAddon = "OBJECTNAME" uses the object name, ie the root prim

Setting  g_sAddon = "something else"; uses "something else" as addon name with no change

Line 37:
```
string g_sAddon = "PRIMNAME"; //  "OBJECTNAME" : replaced by object name, "PRIMNAME" : replace by prim name
```

Lines 428/429:
```
        if (g_sAddon == "OBJECTNAME") g_sAddon = llGetLinkName(LINK_ROOT);
        if (g_sAddon == "PRIMNAME") g_sAddon = llGetLinkName(LINK_THIS);
```

Medea-Destiny (approved):

Surssin:
So I tried it this morning, and the leash seemed to only go from the root prim. I tried to rename the link (P.S. your instructions have a typo, "leaspoint" instead of "leashpoint"), tried to put the script in the root and just the link I wanted it to come from, still the same.

In case it matters, the collar has parts named like a normal OC collar. I don't know if that would confuse the script or not, but changing the names of the links didn't seem to matter.

NikkiLacrima:
Hi, Thanks for testing, Try to first deactivate and then reset , and reactivate the script after renaming link prims. When renaming linked prims It is easy to end up with nonattached chains.  a simple  script with   
llLinkParticleSystem(LINK_SET,[]); 
helps then. Unfortunately the changed event does not detect name change for linked prims

Surssin:
That did seem to do it. Thank you. Sorry if I missed that instruction in the text.
Unless I missed it, it might be nice to have an option where if you touch the object with the relay, it doesn't bring up the menu at all. I realize I may be a niche customer there, but in my case, it's better to have only one menu pop up, not two. ;) But the script is mod, so I was also able to just parse out the touch event without breaking it. ;)

NikkiLacrima:
Nice that it works, you can set the object to ignore touches in  object editor /Edit/General/Click to: "Ignore Object"

Surssin:
Wouldn't work if I want the object to respond to touches otherwise.
As I said, I'm a niche case. I'm using it for my feral form, which has its own collar and gag system, but I like having the OC leash jump to that collar if someone is holding it.

PR #1173, opened 2025-11-10
RLVsuite bitmask correction values

Medea-Destiny:
Updated bitshift calculations were not returning correct values for second byte in restrictions due to correction values (restrictions list is not full byte sized). Fixed correction values in multiple places, so both listing AND setting of restrictions stored in the second byte (corresponding to values >=93 in g_lRLVList) should now both display and set correctly.

### DIFF
--- a/src/collar/oc_rlvsuite.lsl
+++ b/src/collar/oc_rlvsuite.lsl
@@ -54,8 +54,7 @@ medea (medea destiny)
                     with RLV menu
                 -   Removed double menu for (prefix) restrictions
       Jun 2024  -   Extended chat command function above to allow for capitalization of individual restriction names.
-      Nov 2025  -   Fixed correction values for second byte of restrictions to go with bitshift changes in various 
-                    places.
+      Nov 2025  -   Changed macros setting to rlvsys_macrosnew and added conversion function.
 Nikki Lacrima
       Sept 2024 -   Remove superflous llOwnerSay commands and replace g_lMenuIDs (Yosty patch PR #963)
       Oct  2024 -   Correct off by one bitshift and improve bitpos and bitvalue handling for restriction bitmaps. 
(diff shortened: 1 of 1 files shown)


NikkiLacrima:
This will work but what was the issue before ?? Can you give an example where the previous code breaks ?

There are 31 entries in each half placed at positions 0 to 3*30=90, and 93 to 183 in g_lRLVList.
With this patch the first entry in the second half,     
"Add Attach"    , 8 , "addattach"                           ,    // 31  0   1
is mapped to the second bit in the second bitmask, not to the first. Why not map it to the first bit, value 0, in the second bitmask as it was ??

Medea-Destiny:
@NikkiLacrima The issue was with conversion of bitmap to command via g_iRLVList for the second byte, which was being offset by one place. Thus for example, the dress restriction which applies bits 0,1,2,3 of the second byte was in fact issuing remattach, addoutfit, remoutfit and notecard (bits 1,2,3 and 4). 

I'm not sure why these decisions were made with the original implementation. I would suggest a redo to use the full bitmap would be a good choice at some point. We could just insert a blank stride in g_iRLVList (or add a currently unsupported function) to streamline things more easily. This fix simply corrects for the current list values.

Medea-Destiny:
Ok, so Nikki's code is good and yet in my collar it's not working. This is because in the past I have changed and restored the macros, which means the old values are store in oc_settings and overwriting the new one. As Nikki says this patch is therefore incorrect and the patch should be withdrawn. However it also means that anyone who's edited macros in the past is going to get broken macros. 

Now we could deprecate rlvsys_macros in the installer, though currently the installer system only has setting deprecation hard-coded. That's not a great idea though would be easy to fix, though we're going to need a bit of an updater re-write anyway for 9.x so not really needed right now.

HOWEVER, deleting the current settings would mean that anyone who has custom macros would lose them. Instead we could switch to a different token for the macros, and update the old tokens as follows:

           ```
 else if (llList2String(lParams, 0) == "rlvsuite_macros") 
           {
                list temp = llParseStringKeepNulls(llList2String(lParams, 1), ["^"],[]);
                integer i=llGetListLength(temp)-1;
                while(i>1)
                {
                    temp=llListReplaceList(temp,[llList2Integer(temp,i)>>1],i,i);
                    i=i-3;
                }
                llMessageLinked(LINK_THIS, LM_SETTING_SAVE, "rlvsuite_macrosnew=" + llDumpList2String(temp,"^"), "");
                llMessageLinked(LINK_THIS,LM_SETTING_DELETE,"rlvsuite_macros",""); 
           }
```

I have switched the patch over with this code, which should resolve the issue without anyone losing current custom macros.

NikkiLacrima (approved):
Tested macro conversion from 8.3 okay.
Tested restrictions in use, no issues found.
Code looks good.

Approved

PR #1175, opened 2025-11-14
Issue setrot locally rather than through link message to oc_rlvsys

Medea-Destiny:
Issue setrot commands locally rather than through RLVsys. When turnmode is on, you get a lot of setrot commands being set. These are one-shot functions that don't impact any other RLV stuff, so it makes sense to fire it off locally rather than sending a link message that triggers every script constantly while leashed.

### DIFF
--- a/src/collar/oc_particle.lsl
+++ b/src/collar/oc_particle.lsl
@@ -1,173 +1,162 @@
+// This file is part of OpenCollar.
+// Copyright (c) 2008 - 2023 Lulu Pink, Nandana Singh, Garvin Twine,
+// Cleo Collins, Satomi Ahn, Joy Stipe, Wendy Starfall, Romka Swallowtail,
+// lillith xue, littlemousy, Nikki Lacrima et al.
+// Licensed under the GPLv2.  See LICENSE for full details.
+/*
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
ok! Modified significantly further to include:
 (optional) use of llSetAgentRot instead of RLV @setrot for turning avatar. This is triggered by "LSL turn" in the configure menu.

"Face" command. Menu turn opens listing of objects and avatars, when selected wearer will turn to face them. Chat commands "face" opens menu, "face me" turns avatar to face command issuer, "face (string)" faces found object/avatar who's name starts with (string).

Added strict(ish) mode. This is strict mode but without fartouch restriction, for people who want strict leashing to control movement, but not other world interactions. Clicking the "strict" button in configure toggles through the modes in the order off, on, ish. This order ensures that current settings retain current functionality.

GeminiMarshdevil:
i tested this with medea, all is good and fancy and beautiful and frothy bubbly golden juicy. 

did i do good mistress? :v

SilkieSabra:
I took Gem's word for it, Gem should have approving review status in here.

PR #1176, opened 2025-11-15
Modify particle scale in installer

Medea-Destiny:
Particle scale is holdover from before particles were collar shaped. Smaller particles (as used in current 8.4 alphas) IMO look a bit better, less chaotic, preserve rainbow effect better.


Lucifer-Baphomet:
I like the particle size. It's a lot clearer and easier to make out what it is. I used to always think it was a big fucking rainbow snake.

I prefer The Voice. The Voice is sort of sexy in a 1970s science fiction way. I quite enjoy it.

SilkieSabra (approved):

PR #1181, opened 2025-12-20
Leash fix number 4183945

Medea-Destiny:
Clover spotted a situation with the line-of-sight thing in long term testing where strict mode restrictions were being dropped instantly if the leashee leaves range while the collar is in the state where waiting for line of sight to be restored. 

This change set g_iAwayCounter to manage  to trigger the countdown when line-of-sight is dropped, which should fix the issue. Initial reports are good but let's see how this behaviour pans out in testing.


Medea-Destiny:
Added fix for non-functioning back button in face menu, reported in beta testing

PR #1185, opened 2026-01-27
Delete src/collar/oc_bell.lsl

Medea-Destiny:
Delete bell from src/collar because the new version is moved to apps as it's an app not a core file


NikkiLacrima (approved):
Approved. Failed linting checks is false alarm

PR #1186, opened 2026-01-27
Update script version to 8.4

Medea-Destiny:
Updating script versions to match release


NikkiLacrima (approved):
Approved. Failed linting checks is false alarm

PR #1195, opened 2026-03-04
Add @accepttprequest to Exceptions

PhenomeNeon:
This PR adds `@accepttprequest` to the Exceptions list, allowing the automatic acceptance of teleport requests sent by the owner (default - matching `@accepttp`) or trusted users.

### DIFF
--- a/src/collar/oc_rlvextension.lsl
+++ b/src/collar/oc_rlvextension.lsl
@@ -152,7 +152,8 @@ list lRLVEx = [
     "RcvEmote"      , "recvemote"   , 8     ,
     "Lure"          , "tplure"      , 16    ,
     "Force TP"      , "accepttp"    , 32    ,
-    "Start IM"      , "startim"     , 64
+    "Start IM"      , "startim"     , 64    ,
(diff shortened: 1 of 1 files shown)


Medea-Destiny:
Interesting suggestion! This change adds a accepttprequest as the high bit value in our exceptions mask. If set, this would allow an owner to send a tp request to a sub and the sub's viewer will automatically grant the request to teleport to them, as the inverse of the current forceTP option. 

@PhenomeNeon 's script names this function "TP Request" which I'm not sure is as clear as it could be as to what it does. Any thoughts?

PhenomeNeon:
Oops! I just noticed a mistake. If you look at the edit to the comment, you can see I was also not sure about the name. Since I edited the comment last, I think "**Accept TP**" is what I actually wanted to call it. 

But.. interested in others' thoughts before I make a new commit, because I'm still not sure.

NikkiLacrima (approved):
Code looks good and tested basic funtionality is ok. No problems found. 
Name "TP Request" is consistent with Firestorm teleport menu "Request Teleport"

PR #1208, opened 2026-05-16
Update lslint.yml

NikkiLacrima:
Use https://github.com/NikkiLacrima/lslint with updated definitions from https://github.com/secondlife/lsl-definitions


SilkieSabra (approved):
Cool beans, hope this works.

PR #1217, opened 2026-07-02
Tiny performance tuning in CalcAuth

cruelraremaster:
Tiny performance tuning:
`llListFindList(g_lBlock,[sID])` is executed in any cases of the if under any circumstances. In some scenarios it is executed twice - so the line

   `if(llListFindList(g_lBlock,[sID])!=-1)return CMD_BLOCKED;`

could also be executed prior the if, which provides a tiny bit of performance e.g. in cases somebody being neither wearer nor owner clicks the collar.

### DIFF
--- a/src/collar/oc_api.lsl
+++ b/src/collar/oc_api.lsl
@@ -152,10 +152,10 @@ integer CalcAuth(key kID) {
     if(llLinksetDataRead("auth_WearerLockout")!="" && kID==g_kWearer) return CMD_NOACCESS;       
     string sID = (string)kID;
     // First check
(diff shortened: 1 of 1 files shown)


NikkiLacrima:
Looks correct, if kID is blocked it will always return  CMD_BLOCKED  irrespective of other conditions, so it can be move up. Still should be given sanity testing, not that the performance will ever be noted, but the logic is clearer.

NikkiLacrima (commented):
Looks good, approved.

NikkiLacrima (approved):

PR #1218, opened 2026-07-13
Update Makefile for lslint v1.3

cruelraremaster:
The v1.0.8 release of lslint is a binary of type  `Mach-O executable i386`
which means it does not runs on a 64bit macOS, especially on 32bit Intel CPU. Apple moved to "64bit-only" with the release of macOS Catalina  in 2019. Later in 2020 Apple move to the Apple Silicon M CPUs, with M1 being introduced in 2020.

So basically lslint version v1.0.8 does not longer work on macOS development hardware since 2020.

Latest official release is v1.3 ( https://github.com/Makopo/lslint/releases#release-v1.3 ), which is of type `Mach-O 64-bit executable x86_64` which at least is executable on a Apple M2 CPU via the Apple Rosetta2 x86_64-on-Silicon emulation.

Nevertheless, this needs to be testet for compatibility especially on Linux/Win dev systems !
One might want at least to delete the bin/lslint and execute make:

>    $ rm -rf bin
>    $ make
>    mkdir -p bin
>    cd bin; curl -L https://github.com/Makopo/lslint/releases/download/v1.3/lslint_v1.3_osx.zip -o lslint_v1.3_osx.zip
>    cd bin; unzip lslint_v1.3_osx.zip
>    Archive:  lslint_v1.3_osx.zip
>      inflating: lslint
>    $ bin/lslint --version
>    lslint v1.3

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ else
 	endif
 endif
 
-LSLINT_VERSION := v1.0.8
+LSLINT_VERSION := v1.3
 LSLINT_FILE := lslint_$(LSLINT_VERSION)_$(LSLINT_ENV).zip
 LSLINT_URL := https://github.com/Makopo/lslint/releases/download/$(LSLINT_VERSION)/$(LSLINT_FILE)
 LSLINT := bin/lslint


NikkiLacrima:
Updating the version is good thing, the problem we face is that Makopo lslint repo uses old builtins.txt that doesnt contain definitions for GLTF (PBR) materials and other new functions.

The workaround for this is, for now, is to fork lslint into  https://github.com/NikkiLacrima/lslint and manually update the builtins,txt  from https://github.com/secondlife/lsl-definitions/tree/main/generated/builtins.txt. This works well for checking pull requests to OpenCollar controlled by our LSlint action workflow. But I assume when running make we will bring down the old Makopo version and linting will fail. 

So is there a way to update the Makefile to build the lslint binary in similar fashion as for the linting workflow https://github.com/OpenCollarTeam/OpenCollar/blob/main/.github/workflows/lslint.yml  or simply stop linting in the makefile and rely on having linting done as part of pull requests.

